PR #49, opened 2022-09-13
feat: Make express types less strict

netiperher:
Allow using both `express()` and `Router()` objects for `createExpressEndpoints`

### DIFF
--- a/.changeset/sweet-trains-cover.md
+++ b/.changeset/sweet-trains-cover.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/express': minor
+---
+
+Make createExpressEndpoints take IRouter instead of Express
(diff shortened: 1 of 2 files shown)


.changeset/sweet-trains-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ts-rest/express': major
```
oliverbutler, line 2: If this is a backwards-compatible update, let's make this a minor update :)

oliverbutler (changes requested):
Can we make this minor? Thanks for the PR!

.changeset/sweet-trains-cover.md
netiperher: good point, fixed in new version.

netiperher:
thanks for reviewing this, I've pushed a new version now

oliverbutler (approved):

oliverbutler:
@all-contributors please add @netiperher for docs, code

PR #53, opened 2022-09-14
feat: Set operationId in open-api docs

netiperher:
From what I understand the route keys in the contract should be unique within the router so they are suitable candidates for using as the operationId when generate openapi docs.

The only thing I'm not 100% sure about is when using nested routers, I guess there is no easy way to ensure strict uniques throughout the whole contract in that case?

### DIFF
--- a/.changeset/red-seals-fry.md
+++ b/.changeset/red-seals-fry.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/open-api': minor
+---
+
+Allow setting operationId in open-api docs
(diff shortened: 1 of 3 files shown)


oliverbutler:
Hmmm... interesting, I don't have masses of context with OpenAPI but I wanted to get the ball rolling with this lib.

Looking at the Swagger docs on https://swagger.io/docs/specification/paths-and-operations/, it looks like as you've said, nested routes could pose an issue for uniqueness.

Potentially, we could run a check at runtime, whilst initializing OpenAPI which would check the uniqueness of the keys and provide a warning/error. 🤔

netiperher:
I've added a runtime option for controlling this behaviour now together with an error, so let me know what you think about that approach.

oliverbutler (approved):
Nice! This looks great and has been done in a non-breaking way, I might spend some time on some documentation this weekend. I'll try to document this behaviour in the docs.

PR #60, opened 2022-09-19
Migrate to pnpm and add microservice example

oliverbutler:


oliverbutler:
Migrated to PNPM and setup pipeline caching (solves #63)

PR #73, opened 2022-11-16
feat: update custom api docs

michaelangrivera:
Hi! 

Figured I'd try to help out and contribute back. 

It doesn't seem `baseUrl` is accessible in the `api` method. It may be missing as a type, or not a feature yet. I can look into that and open a PR. 

I moved all the custom stuff to a `Custom Client` page to make it more apparent since I feel there will be many other people wanting to take advantage of this. 

Thank you, and let me know what you think!


oliverbutler:
@michaelangrivera This looks perfect, one way our company has navigated the issue of updating baseUrl and headers at runtime is to get and set the auth tokens and URL in the API callback, without setting a baseUrl at all.

You're more than welcome to open a PR, I will happily assist, it would be _super_ appreciated!

arivera-xealth:
@oliverbutler Awesome thank you! Let me know when the new docs are deployed!

oliverbutler:
@all-contributors please add @michaelangrivera for docs

PR #78, opened 2022-12-08
feat: add credentials option for fetch client

Gabrola:
For APIs authenticated using cookies, the cookies are not being sent by default for cross-origin requests, as the default behavior for the fetch API is to use a `same-origin` policy. This means if the API is even on a different subdomain (api.example.com) or a different port, say when running locally, no cookies will be sent. This happens regardless of any CORS headers sent by the backend API.

The only way to solve this problem is to use a custom client using the `api` option, so I've added a `credentials` option that will set the same option on the Fetch API to avoid overriding the client.

I've documented this feature in the docs and added the missing sidebar link to the "custom client" page.

### DIFF
--- a/.changeset/witty-apricots-exercise.md
+++ b/.changeset/witty-apricots-exercise.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/core': minor
+---
+
+Add credentials option for fetch client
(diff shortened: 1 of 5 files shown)


oliverbutler:
LGTM!

This looks awesome, thanks so much @Gabrola! This should help a lot of people, I appreciate the PR hugely.

PR #81, opened 2022-12-10
Improved query string encoding and decoding

Gabrola:
Following up on https://github.com/ts-rest/ts-rest/pull/80, I've transitioned the nest library to use `req.query` and `req.params` rather than parsing them from the URL. This now allows the usage of arrays and nested objects in query params. I've made sure that this use-case is also covered in the clients by changing `convertQueryParamsToUrlString` to use `qs.stringify` which is the same library used internally by express.js.

Also tweaked the Nest implementation such that the all contracts' AppRoutes are added to their respective methods' metadata. This happens only once during runtime at startup, and therefore does need to be added to each request object. This allows us to maintain a single instance of `ApiRouteInterceptor` in the IoC container across all routes.

I've also changed how the status and response body are returned from `ApiRouteInterceptor`. Throwing an `HttpException` meant that no other interceptors could have been used since the thrown exception goes straight to Nest's base exception filter.

Added some tests in the example-nest app.


oliverbutler:
Hey @Gabrola! I was discussing how we could improve the query params parsing whilst chatting with a Colleague yesterday, so this is amazing!

My only concern was the added peer dependency, and whether it was worth the extra [9.8kB (Gz) requirement](https://bundlephobia.com/package/qs@6.11.0) for qs (especially in the core package)

Do you think this is a worthy requirement? 🤔

oliverbutler:
Thinking on this, as we should support Nest and Express with as few dependencies as possible is it worthwhile to keep using encodeURIComponent in the client(s), then qs or Nest/Express specific implementations on the server libs?

qs is not likely to be an existing dep on many frontends right?

Gabrola:
Thank you for your review! Yes, you're absolutely correct. qs is already a dependency of express/nest for its `parse` function, but it seems like the `convertQueryParamsToUrlString` function is only used by the frontend libraries. I agree though that it seems too large for what we need from it.

It should be straightforward to implement what we need, since qs has too many options (and code) to cater to other options such as comma-delimited arrays, etc.

Which makes me think, do you think it makes sense to add an option client-side to plug in a custom query serializer? Generally on the frontend you would be able to do this easily by passing your query params as string directly to fetch/got/axios, etc. However, by serializing in ts-rest we are essentially locking everyone in to our kinda-opinionated serialization.

There are some cases I could think of that might want to be handled differently in different apps. For example, in case of null do we serialize as `?foo=null`, or just `?foo`, or do we omit it entirely? What about undefined?

If we provide the option of passing a custom serializer (or qs 😆), we'd have the best of both worlds of not requiring an extra dependency with the core library, while also giving developers the flexibility if needed. Seems fitting now, since on the backend we are directly consuming `req.query`, so its already a hands-off experience when it comes to deserialization. But now the point raised in https://github.com/ts-rest/ts-rest/pull/80 comes into contention, I think now it'd make sense to allow custom `style` and `explode` options for query serialization in OpenAPI.

oliverbutler:
I like your suggestion.

## Proposed Client `queryParamsEncode`

```typescript

// qs
export const postsClient = initQueryClient(postsApi, {
  ...
  queryParamsEncode: ({params, route}) => `?${qs.stringify(params)}`
});

// JSON (or any other preferred encoding?)
export const postsClient = initQueryClient(postsApi, {
  ...
  queryParamsEncode: ({params, route}) => `?json=${JSON.stringify(params)}`
});
```

If we did this with the client it would allow any _general_ encoding to be applied to a query client

In terms of standardisation, it's quite messy overall... doing this approach could allow ts-rest users to work with third-party API, no matter how they encode their params I suppose.


## Proposed new default behaviour behaviour
My next question is what _we_ standardise on, as the default between ts-rest clients and server implementations.

I don't like the unintuitive behaviour given that right now arrays don't function correctly...

### Implement our own formatted query string (OpenAPI standard)

Can probably do this if we limit the scope to `explode` and `deepObject`

As can be seen here, Nest.js and Express happily parse these.

Screenshot 2022-12-10 at 14 40 30

There is still a problem here, we _still_ need to use Zod or similar to parse strings into numbers, enums etc. 🤔

oliverbutler:
@Gabrola, this was deceivingly simple... I'll keep trying some more unit tests.

Screenshot 2022-12-10 at 15 10 26

oliverbutler:
cc @joshjacks this may help what we discussed last week 🤔

oliverbutler:
@Gabrola I just checked every imaginable data shape, nested structure, array, array in an array, dates, null, undefined etc and landed on this solution

```typescript
/**
 *
 * @param query - Any JSON object
 * @returns - The query url segment, using explode array syntax, and deep object syntax
 */
export const convertQueryParamsToUrlString = (
  query: Record<string, unknown>
) => {
  const queryString = encodeQueryParams(query);
  return queryString?.length > 0 ? '?' + queryString : '';
};

export const encodeQueryParams = (query: Record<string, unknown>) => {
  return Object.keys(query)
    .flatMap((key) => tokeniseValue(key, query[key]))
    .map((pair) => {
      const [key, ...rhs] = pair.split('=');
      return `${key}=${rhs.map(encodeURIComponent).join('=')}`;
    })
    .join('&');
};

const tokeniseValue = (key: string, value: unknown): string[] => {
  if (Array.isArray(value)) {
    return value.flatMap((v, idx) => tokeniseValue(`${key}[${idx}]`, v));
  }

  if (value instanceof Date) {
    return [`${key}=${value.toISOString()}`];
  }

  if (value === null) {
    return [`${key}=`];
  }

  if (value === undefined) {
    return [];
  }

  if (typeof value === 'object') {
    return Object.keys(value).flatMap((k) =>
      // @ts-expect-error - accessing object keys
      tokeniseValue(`${key}[${k}]`, value[k])
    );
  }

  return [`${key}=${value}`];
};
```

This is considerably better than a 10KB dependency on the client (I think!), hopefully, we can then just utilize the express/nest built-in parsers.

Also for every test I ran the resulting query string though qs, to make sure it was able to parse it happily.

## Note!

There is one difference, I went with the `?foo[0][0]=bar&foo[0][1]` syntax to achieve nested arrays, fully qs compatible I believe.

Gabrola:
@oliverbutler that's pretty neat! That's a much simpler than what I was working on since I was using the `?foo=bar&foo=baz` format for arrays and that brought a bit of complexity. I'll add your implementation in with the extra unit tests I wrote and let me know what you think

oliverbutler:
@Gabrola I'm going to push this change to main, to be honest, it's not breaking as right now anyone who used an array/object definitely got runtime errors 

Plus it's already got all of the unit tests, once that's done maybe just rebase from main? :)

oliverbutler:
* wait correction, I can just push to your branch so we can polish this.

Gabrola:
Concerning doing zod validations, I've messed a bit with it and finally landed on this solution.

```typescript
const stringNumberTransform = (val: string, ctx: z.RefinementCtx) => {
  const parsed = parseInt(val, 10);
  if (Number.isNaN(parsed)) {
    ctx.addIssue({
      code: z.ZodIssueCode.custom,
      message: 'Not a number',
    });
    return z.NEVER;
  }
  return parsed;
};

  findAll: {
    method: 'GET',
    path: '/posts',
    summary: 'Get a paginated list of posts',
    query: z.object({
      page: z.string().default('1').transform(stringNumberTransform),
      pageSize: z.string().default('50').transform(stringNumberTransform),
    }),
    responses: {
      200: c.response<Post>(),
    },
  },
});
```

This gives us an input type of 

```typescript
{
    page?: string | undefined;
    pageSize?: string | undefined;
}
```

and an output type of 
```typescript
{
    sort: "asc" | "desc";
    sortBy: "name" | "createdAt" | "updatedAt";
    page: number;
    pageSize: number;
}
```

The recommended solution by zod is to use `.preprocess()` but that causes the input types to be `unknown`.

oliverbutler:
yeah, this is pretty much identical to how we solved a problem on our side like this... It's not perfect

Gabrola:
@oliverbutler sure, yeah that sounds better if you push onto this branch. I pushed my changes to it so far, and split the query functionality into its own file.

oliverbutler:
@Gabrola Pushed! 👍 Commented out the query.ts export temporarily so you can reconcile the two implementations

Gabrola:
@oliverbutler Perfect, thanks!

oliverbutler:
@Gabrola I _think_ I tested pretty much everything, but if there are any oversights happy to help out.

Gabrola:
@oliverbutler Your code passes all of the unit tests I thought of as well so it seems like everything's good. Only thing is that it seems like square brackets in the keys need to be URL encoded too, so I'll make sure that it done too.

oliverbutler:
Btw @Gabrola you've absolutely helped smash some awesome improvements out, I decided to add the allcontributors.org bot! Hopefully, you'll appear in the README :) 

I really appreciate the help and ideas!

Gabrola:
@oliverbutler So glad to hear that! I appreciate it mate!

libs/ts-rest/nest/src/lib/api.decorator.ts
```diff
@@ -6,24 +6,26 @@ import {
...
   AppRoute,
   AppRouteMutation,
   checkZodSchema,
-  getPathParamsFromUrl,
   PathParamsWithCustomValidators,
   Without,
   ZodInferOrType,
 } from '@ts-rest/core';
 import { map, Observable } from 'rxjs';
+import type { Request, Response } from 'express-serve-static-core';
```
oliverbutler, line 26: is this correct in the resulting package.json (not pinned to a super specific version or anything?)

Gabrola: @oliverbutler ideally it shouldn't be added to the package.json at all as it is only being used for typing during compile time, but I just checked it seems like it's being added to the peer dependencies. Let me see if there's a solution for that. For some reason `qs` is being added to core's package.json too despite being used only in a spec file.

Gabrola: The qs problem seems to be an issue with nx https://github.com/nrwl/nx/issues/8096
I recently moved to turborepo because of nx's weird quirks

oliverbutler: Yeah, that's a tad weird. NX should be smart enough to purge it from the resulting json...

Also, I just noticed you removed the `getPathParamsFromUrl` helper, honestly, I can't remember _why_ this existed 🤔 It's probably because I wasn't sure how to access the params as you did!

Maybe we can delete the `libs/ts-rest/core/src/lib/server-utils.ts` in their entirely 🎉

Gabrola: Yeah, let me dig into that nx thing.

As for `getPathParamsFromUrl` it was being used instead of `req.params` in nest, but it looks like it's still being used by next. I guess because the [...ts-rest] param needs to be expanded from a full URL into individual params.

Gabrola: Ok, it should be fine now. Looks like nx isn't able to handle `qs` being imported using a wildcard `import * as qs...` so I've used the named imports instead. For `express-serve-static-core` I've just included it into dev deps so it's not added into the peer deps

oliverbutler: Ok, awesome 🔥 I'll investigate the next thing. I'll see if I can move that code out of the core lib to decrease bundle size.

oliverbutler:
@Gabrola Just pushed the fix to next and core to remove server-utils, is all good feature wise on your side, I'm super happy with this!

Gabrola (commented):
LGTM! Thank you for your help! 🎉

PR #82, opened 2022-12-10
fix: use z.input for body and query types on clients

Gabrola:
For cases where zod input and output types differ such as when using `.default()` or `.transform()`, we need to use the input type on the client-side.

```typescript
const foo = z.object({ bar: z.string().default('bar') });
type input = z.input<typeof foo>; // { bar?: string | undefined }
type output = z.infer<typeof foo>; // { bar: string }
```

### DIFF
--- a/.changeset/eleven-roses-know.md
+++ b/.changeset/eleven-roses-know.md
@@ -1,0 +1,7 @@
+---
+'@ts-rest/core': patch
+'@ts-rest/react-query': patch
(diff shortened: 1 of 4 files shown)


oliverbutler:
Ah! Thank you! I scanned over the docs to look for this method but couldn't find it 🌟 🙌

oliverbutler:
@all-contributors please add @Gabrola for code, ideas, test

PR #94, opened 2022-12-10
fix: let nx generate correct package.json

Gabrola:
NX strikes again 😅
Looks like with the update to version 15.3, it started generating the libraries with a different file structure, and it's causing this error `TS7016: Could not find a declaration file for module '@ts-rest/react-query'.`

Screenshot 2022-12-11 000846

NX automatically generates this if it is not declared explicitly

```json
{
  "module": "./index.js",
  "main": "./index.cjs",
  "type": "module",
  "types": "./src/index.d.ts"
}
```

@ts-rest/react-query is the only package with these paths explicitly declared and became invalid when nx started generating index.d.ts inside ./src/index.d.ts instead of ./index.d.ts

### DIFF
--- a/.changeset/wise-owls-hunt.md
+++ b/.changeset/wise-owls-hunt.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/react-query': patch
+---
+
+Fix missing typings due to new declaration file path
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Create a contract, implement it on your server then consume it in your client. I
 
 ## Stargazers over Time ⭐️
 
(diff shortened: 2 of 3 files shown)


oliverbutler:
Ahh! Good catch, I wonder if somewhere in this repo we should have a _non NX_ app which imports the libs to sanity check them, rather than using NX to import the TS code directly.

Gabrola:
Yeah, same train of thought came to me when I noticed that this wasn't caught by any tooling. I just tested a quick idea in `apps/example-microservice/web-app/src/main.tsx` changing the react-query import into `import { initQueryClient } from '../../../../dist/libs/ts-rest/react-query'`, and that would have errored out before this fix.

PR #75, opened 2022-11-26
chore: fix ESM support

michaelangrivera:
## Context

Hope to officially close #66 without workarounds. To be clear, I _do not_ want to merge these changes just yet, but I want to get at least the conversation going, and a PR started so that we can implement @ecyrbe 's suggestions or our own fix. 

## Background

On my NextJS webpack build, and several others, we observe the following error: 
```
Server Error
Error: No QueryClient set, use QueryClientProvider to set one

This error happened while generating the page. Any console logs will be displayed in the terminal window.
```

Implementing this workaround no longer works for me (for some reason): 

```
    if (options.isServer) {
      config.externals = ["@tanstack/react-query", ...config.externals];
    }
    const reactQuery = path.resolve(require.resolve("@tanstack/react-query"));
    console.log(reactQuery, 'converted ts-rest imports for react-query')
    config.resolve.alias["@tanstack/react-query"] = reactQuery;
    return config;
  },
```

According to #66  and @ecyrbe, the likely culprit is that exports are not being correctly interpreted by webpack. See: https://github.com/ecyrbe/zodios/issues/191#issuecomment-1288101919. 

I released a temporary test version of `@ts-rest/react-query` here: https://www.npmjs.com/package/michaelangrivera-ts-rest-react-query, and I still see the same issue while implementing that suggestion. 

Interestingly, though, the following works:
```
`const client = require('michaelangrivera-ts-rest-react-query')`
```
I guess this is to be expected, though, since ESM is failing and not common modules. 

This is what the build looks like after `nx build` for that library:
```
{
  "name": "michaelangrivera-ts-rest-react-query",
  "version": "3.6.6",
  "description": "react-query client integration for @ts-rest",
  "license": "MIT",
  "main": "./index.cjs",
  "module": "./index.js",
  "typings": "index.d.ts",
  "exports": {
    ".": {
      "types": "./index.d.ts",
      "import": "./index.js",
      "require": "./index.cjs"
    },
    "./package.json": "./package.json"
  },
....
  "peerDependencies": {
    "react": "16.x.x || 17.x.x || 18.x.x",
    "zod": "3.x.x",
    "@tanstack/react-query": "4.x.x",
    "@ts-rest/core": "3.6.0"
  },
  "type": "module",
  "types": "./index.d.ts",
  "dependencies": {}
}
```

I tried the following changes (I published them, re-installed, and cleared .next cache):
-  `"generateExportsField": true,` no luck
- only using `esm` under the `format` configuration, no luck

[...]

PR #96, opened 2022-12-11
fix: revert react-query exports in package.json

Gabrola:
@oliverbutler Sorry about this, I removed the exports from the package.json in the last PR and it seems like it broke something and I have no idea why, but you probably had it in there for the some valid reason. Without the exports in package.json I get this error in Next.js

`No QueryClient set, use QueryClientProvider to set one`

From some quick research it seems like this happens when different react-query versions are being used, but I could not observe that in my case so I dunno honestly 🤷‍♂️

### DIFF
--- a/.changeset/flat-rings-sell.md
+++ b/.changeset/flat-rings-sell.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/react-query': patch
+---
+
+Revert @ts-rest/react-query exports in package.json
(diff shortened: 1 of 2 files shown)


oliverbutler:
Yup I'll take a look thanks, those exports were added for both CJS and ESM compatibility.

We probs just want to move the index.d.ts path :)

oliverbutler:
@Gabrola I just pushed a fix directly for this, I don't want to affect this much considering @michaelangrivera is trying to fix this currently on another PR! https://github.com/ts-rest/ts-rest/pull/75

PR #106, opened 2022-12-16
fix: don't use $ref in OpenAPI

Gabrola:
If the same zod object is used multiple times in the same contract object (query, body or response), it will use a $ref to refer to it's first occurrence. For the example in test code, it would have generated to following JSON schema.

```json
"comments": {
  "anyOf": [
	{
	  "items": {
		"additionalProperties": false,
		"properties": {
		  "id": {
			"type": "number"
		  },
		  "title": {
			"type": "string"
		  }
		},
		"required": [
		  "id",
		  "title"
		],
		"type": "object"
	  },
	  "type": "array"
	},
	{
	  "items": {
		"additionalProperties": false,
		"properties": {
		  "author": {
			"type": "string"
		  },
		  "id": {
			"$ref": "#/definitions/zodObject/properties/comments/anyOf/0/items/properties/id"
		  },
		  "title": {
			"$ref": "#/definitions/zodObject/properties/comments/anyOf/0/items/properties/title"
		  }
		},
		"required": [
		  "id",
		  "title",
		  "author"
		],
		"type": "object"
	  },
	  "type": "array"
	}
  ]
}
```

Since we aren't actually bundling any definitions, these references don't resolve and we end up with an error like this.

Screenshot 2022-12-16 210549

We also fix this by naming and bundling all definitions generated, but it'll be so clunky. We'd end up with definitions named something like commentsGetPostCommentsResponses200

### DIFF
--- a/.changeset/long-snails-hammer.md
+++ b/.changeset/long-snails-hammer.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/open-api': patch
+---
+
+Don't use $ref in OpenAPI
--- a/libs/ts-rest/open-api/src/lib/ts-rest-open-api.spec.ts
+++ b/libs/ts-rest/open-api/src/lib/ts-rest-open-api.spec.ts
@@ -9,6 +9,11 @@ type Post = {
   published: boolean;
 };
 
+const commentSchema = z.object({
+  id: z.number(),
+  title: z.string(),
+});
+
 const postsRouter = c.router({
   getPost: {
     method: 'GET',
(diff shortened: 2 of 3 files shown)


oliverbutler:
Amazing, thanks for the PR @Gabrola!

By the way, I haven't actually written any OpenAPI in prod with ts-rest yet, have you? Can we maybe work on some new docs for it?

I definitely think the docs, especially on the OpenAPI side need some (any) love!

oliverbutler (approved):

Gabrola:
@oliverbutler thanks for the review! We are serving Swagger UI internally but I don't think any of us are using them so far. I'm thinking of serving a pure swagger.json through this plugin though and then importing that into postman, as we're still using that mainly.

I think can whip something up in the docs.

Gabrola:
Also loving the rebrand, good job with that! 🎉

PR #114, opened 2022-12-23
feat: add response shape feature, update docs

michaelangrivera:
Per discussion on Discord, this feature allows users to extract response types from a given Contract. This is particularly useful for backend services where you want to avoid implementing an interface. 

I aimed to make this as non-intrusive as possible! All tests pass. Please read the modified Docs to get a feel for the changes/additions, and then you can look at `client.ts` to see the entry point of the implementation. 

Here's a screenshot of how it works:

Screenshot 2022-12-23 at 1 26 32 PM

### DIFF
--- a/apps/docs/docs/core/core.md
+++ b/apps/docs/docs/core/core.md
@@ -95,3 +95,26 @@ export const contract = c.router({
   },
 });
 ```
+
+## Response Types
+
+If you need to quickly extract the Response types from a given Contract, you can use `getRouteResponses`.
+
+:warning: This does not support nested contracts yet. You need to pass in a defined contract. 
+
+`getRouteResponses(router)` returns an empty object with defined response types for a contract's HTTP method. 
+
+```typescript
+import { getRouteResponses } from '@ts-rest/core'
+import { contract } from './contract'
+
+//  initContract() must called prior
+const contractResponses = getRouteResponses(contract)
+type ResponseShapes = typeof contractResponses
+
+async function someHttpCall(req: Request): Promise<ResponseShapes['getPosts']> {
+  return ...
+}
+```
+
+This is particularly useful for Services/Functions/Lambdas that do not use frameworks like Express or Nestjs. 
--- a/apps/docs/docs/nest.md
+++ b/apps/docs/docs/nest.md
@@ -55,10 +55,10 @@ You have two options to ensure HTTP type safety on your Nest Controllers:
         const post = await this.postService.getPost(id);
 
         if (!post) {
-          return { status: 404 as const, body: null };
+          return { status: 404, body: null };
         }
 
-        return { status: 200 as const, body: post };
+        return { status: 200, body: post };
       }
     }
     ```
--- a/apps/docs/docs/nest.md
+++ b/apps/docs/docs/nest.md
@@ -62,7 +62,7 @@ You have two options to ensure HTTP type safety on your Nest Controllers:
       }
     }
     ```
-  - If your controller needs to implement a difference class, or needs extra methods defined outside of a contract, this is option gives that flexibility
+  - If your controller needs to implement a difference class, or needs extra methods defined outside of a contract, this is option gives that flexibility without having to worry about maintaining class extensions. 
 
 
 :::caution
--- a/apps/docs/docs/nest.md
+++ b/apps/docs/docs/nest.md
@@ -6,6 +6,7 @@ By default Nest doesn't offer a nice way to ensure type safe controllers, primar
 const s = initNestServer(apiBlog);
 type ControllerShape = typeof s.controllerShape;
 type RouteShape = typeof s.routeShapes;
+type ResponseShapes = typeof s.responseShapes;
 
 @Controller()
 export class PostController implements ControllerShape {
@@ -27,6 +28,42 @@ export class PostController implements ControllerShape {
 The `@Api` decorator takes the route, defines the path and method for the controller route.
 
 It also injects "appRoute" into the req object, allowing the `@ApiDecorator` decorator automatically parse and check the query and body parameters.
+
+
+## Response Return Type Safety
+
+You have two options to ensure HTTP type safety on your Nest Controllers: 
+
+- `ControllerShape` as shown above: 
+  - Your controller can implement the `ControllerShape` derived from `typeof s.controllerShape`
+  - This ensures your controller methods also align with the base interface shapes
+- `ResponseShapes`:
+  - Your controller can utilize `typeof s.responseShapes` in the return type. For example:
+    ```typescript
+    const s = initNestServer(apiBlog);
+    type ControllerShape = typeof s.controllerShape;
+    type RouteShape = typeof s.routeShapes;
+    type ResponseShapes = typeof s.responseShapes; // <- http Responses defined in contract
+
+    @Controller()
+    export class PostController {
+      constructor(private readonly postService: PostService) {}
+
+      @Api(s.route.getPost)
+      async getPost(@ApiDecorator() { params: { id } }: RouteShape['getPost']):
+      Promise<ResponseShapes['getPost']> { // <- return type defined here
+        const post = await this.postService.getPost(id);
+
+        if (!post) {
+          return { status: 404 as const, body: null };
+        }
+
+        return { status: 200 as const, body: post };
+      }
+    }
+    ```
+  - If your controller needs to implement a difference class, or needs extra methods defined outside of a contract, this is option gives that flexibility
+
 
 :::caution
 
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -67,6 +67,20 @@ export type ApiRouteResponse<T> =
       body: unknown;
     };
 
+export type ApiResponseForRoute<T extends AppRoute> =
+  | {
+      [K in keyof T['responses']]: {
+        status: K
+        body: ZodInferOrType<T['responses'][K]>
+      }
+    }[keyof T['responses']]
+  & {
+      status: Exclude<HTTPStatusCode, keyof T>
+      body: unknown
+    };
+
+
+
 /**
  * Returned from a mutation or query call
  */
@@ -225,3 +239,13 @@ export const initClient = <T extends AppRouter>(
 
   return proxy as InitClientReturn<T>;
 };
+
+// takes a router and returns response types for each AppRoute
+// does not support nested routers, yet
+
+export function getRouteResponses<T extends AppRouter>(router: T) {
+  return {} as {
+     [K in keyof typeof router]: 
+        typeof router[K] extends AppRoute ? ApiResponseForRoute<typeof router[K]> : 'not a route'
+   }
+}
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -67,18 +67,7 @@ export type ApiRouteResponse<T> =
       body: unknown;
     };
 
-export type ApiResponseForRoute<T extends AppRoute> =
-  | {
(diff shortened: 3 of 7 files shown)


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -67,6 +67,20 @@ export type ApiRouteResponse<T> =
...
+      [K in keyof T['responses']]: {
+        status: K
+        body: ZodInferOrType<T['responses'][K]>
+      }
+    }[keyof T['responses']]
+  & {
+      status: Exclude<HTTPStatusCode, keyof T>
+      body: unknown
+    };
+
```
michaelangrivera, line 81: I believe the `Intersection Type` makes more sense for response types and the current implementation.

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -225,3 +239,13 @@ export const initClient = <T extends AppRouter>(
...
+// takes a router and returns response types for each AppRoute
+// does not support nested routers, yet
+
+export function getRouteResponses<T extends AppRouter>(router: T) {
+  return {} as {
+     [K in keyof typeof router]: 
+        typeof router[K] extends AppRoute ? ApiResponseForRoute<typeof router[K]> : 'not a route'
+   }
+}
```
michaelangrivera, line 251: Not sure how this is testable, tbh. It's a TS feature, not a runtime feature.

michaelangrivera: I also tried a recursive implementation, but it broke the type-checking. Happy to see if you have any ideas

oliverbutler:
@michaelangrivera Hey just popping in for a bit of visibility, I've been celebrating the holidays the last few days, I will take a loom tomorrow, but at first glance, this looks great!

apps/docs/docs/nest.md
```diff
@@ -28,6 +29,42 @@ The `@Api` decorator takes the route, defines the path and method for the contro
...
+      @Api(s.route.getPost)
+      async getPost(@ApiDecorator() { params: { id } }: RouteShape['getPost']):
+      Promise<ResponseShapes['getPost']> { // <- return type defined here
+        const post = await this.postService.getPost(id);
+
+        if (!post) {
+          return { status: 404 as const, body: null };
+        }
+
+        return { status: 200 as const, body: post };
```
oliverbutler, line 61: is the as const still needed with the return type? :(

libs/ts-rest/core/src/lib/client.ts
oliverbutler: Yeah, I'll pull and try to fix the recursive issue!

apps/docs/docs/core/core.md
```diff
@@ -95,3 +95,26 @@ export const contract = c.router({
...
+
+`getRouteResponses(router)` returns an empty object with defined response types for a contract's HTTP method. 
+
+```typescript
+import { getRouteResponses } from '@ts-rest/core'
+import { contract } from './contract'
+
+//  initContract() must called prior
+const contractResponses = getRouteResponses(contract)
+type ResponseShapes = typeof contractResponses
```
oliverbutler, line 113: Quick question, why is this feature implemented as a function rather than just a generic type?

I saw you already made `ApiResponseForRoute`, we could just make a mapped type right? e.g. this could be a resulting type

```typescript

type Test = ResponseShapes<typeof MyAppRouter>
//       ^?
// {
//  "getPosts" : ApiResponseForRoute<...>
// }
```

oliverbutler: Because you call the function, which just returns {}, then you only use it at compile time with the typeof, feels a bit superfluous maybe?

apps/docs/docs/core/core.md
```diff
@@ -95,3 +95,26 @@ export const contract = c.router({
...
+
+async function someHttpCall(req: Request): Promise<ResponseShapes['getPosts']> {
+  return ...
+}
+```
+
+This is particularly useful for Services/Functions/Lambdas that do not use frameworks like Express or Nestjs.
```
oliverbutler, line 120: 🙌

apps/docs/docs/core/core.md
oliverbutler: Or even a generic that could be used like

```
async getAThing(): RouteResponse<MyApi["getAThing"]> 
```

Or something where you could pass in any `AppRoute`, not 100% sure how I prefer this, lots of good options

michaelangrivera: So I implemented a runtime function to allow the user to use this on any runtime `router` already defined. For instance, if you initialize `router()` already and want to extract the response types somewhere else in your code, you could use this generic function to pull those types out at any time. The function is also used in the `ts-rest-nest.ts` implementation here: 
```
  return {
    controllerShape: {} as NestControllerShapeFromAppRouter<T>,
    routeShapes: {} as NestAppRouteShape<T>,
    responseShapes: getRouteResponses(router), <---
    route: router,
  };
```
I agree we could use a type, but I figured it's a better DX to do something like
`getRouteResponses(router)` versus `ApiResponseForRoute<type of router>` in your code. 

Regarding the generic type, I was limiting it to just response types to limit scope creep/major changes to the core.

apps/docs/docs/nest.md
michaelangrivera: You're right! Even better 🥳 
I updated the example

apps/docs/docs/nest.md
```diff
@@ -28,6 +29,42 @@ The `@Api` decorator takes the route, defines the path and method for the contro
...
+
+        if (!post) {
+          return { status: 404, body: null };
+        }
+
+        return { status: 200, body: post };
+      }
+    }
+    ```
+  - If your controller needs to implement a difference class, or needs extra methods defined outside of a contract, this is option gives that flexibility
```
Gabrola, line 65: I believe this is not entirely accurate. You should be able able to define extra methods other than than ones in the contract and typescript would not complain. You can also implement multiple interfaces or extend another class and also still implement the interface, i.e., `class Controller extends BaseController implements ControllerShape, AnotherInterface`

The only case I could see this useful if you are not fully implementing the entire contract in a single controller. But even then, you can use Pick/Omit to achieve the same goal, so something like `class Controller implements Pick<ControllerShape, 'getPost'>`

I'm not sure if it's worth introducing this change to the Nest plugin as it could be a bit confusing as to what approach should be taken by someone reading the documentation. Also the explicit typing in the method declaration is quite unwieldy

`async getPost(@ApiDecorator() { params: { id } }: RouteShape['getPost']): Promise<ResponseShapes['getPost']>`

Gabrola:
This looks pretty good! I'm not sure about the Nest part of it (see https://github.com/ts-rest/ts-rest/pull/114#discussion_r1058728842), but the generic implementation in `@ts-rest/core` could be pretty useful for other frameworks/backends other than the supported ones.

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -67,6 +67,20 @@ export type ApiRouteResponse<T> =
...
+  | {
+      [K in keyof T['responses']]: {
+        status: K
+        body: ZodInferOrType<T['responses'][K]>
+      }
+    }[keyof T['responses']]
+  & {
+      status: Exclude<HTTPStatusCode, keyof T>
+      body: unknown
+    };
```
Gabrola, line 80: You could simplify this into

```suggestion
export type ApiResponseForRoute<T extends AppRoute> = ApiRouteResponse<T['responses']>;
```

Also an intersection type would cause this to resolve into `never`

apps/docs/docs/nest.md
michaelangrivera:
```
You can also implement multiple interfaces or extend another class and also still implement the interface, i.e., class Controller extends BaseController implements ControllerShape, AnotherInterface
```
This is true, I can update this verbiage. However, overall, observing the specific TS error you're getting with a class-based approach is exponentially difficult. For example, take the `posts.controller` file:

Screenshot 2022-12-29 at 8 10 02 AM

To see the exact TS error you're receiving, you have to do quite a bit of scrolling to see what's wrong.  Over a significantly sized controller, this approach, in my opinion, isn't scalable. Extending a class and finding graining with `class Controller implements Pick<ControllerShape, 'getPost'` is also quite a bit of extra work and not as intuitive in my opinion. It also promotes extra inheritance patterns, which, as we all know, can get quite messy (since TS allows for 3rd level extensions/inheritance). 

```
I'm not sure if it's worth introducing this change to the Nest plugin as it could be a bit confusing as to what approach should be taken by someone reading the documentation. Also the explicit typing in the method declaration is quite unwieldy
```
Again, this is all preference/opinion, which is why it's more of an addition than a change. If you'd both like, I can mark the title in the docs`Optional` for this approach. I know some users would definitely appreciate this feature, including my company.

Gabrola: Aha! Yeah, that totally makes sense now, I definitely agree! Now that I'm checking, it's pretty much the same issue with all of the server libraries.

libs/ts-rest/core/src/lib/client.ts
michaelangrivera: @Gabrola so I tried both types (intersection and union). Let's take an example from `getPosts` contract schema: 
```
    responses: {
      200: z.object({
        posts: PostSchema.array(),
        count: z.number(),
        skip: z.number(),
        take: z.number(),
      }),
    },
```
when using the `union` type, if I were to add an extra key to this response object like so:

✅ 
```
    return {
      status: 200,
      body: { posts, count: totalPosts, skip, take, extraParam: 'hello' },
    };
```
I do not receive a TS error. Using the `intersection` type will result in an error:

```
    return {
      status: 200,
      body: { posts, count: totalPosts, skip, take, extraParam: 'hello'  <--- ❌  },
    };
```

Maybe we should discuss the intended functionality, but in my opinion, the return types should be strict: "Strict" meaning we don't allow extra properties in response object other than what's defined in the contract. Also, if properties are missing that are required, we should also enforce a TS error.

Gabrola: The reason this is happening is because you have a slight bug in your code. The second part was causing it to resolve to a `body: unknown` even for status 200, so you were ending up with a body type of `body: z.infer<...> | unknown` which is why it was letting through extra properties. If you fix that bug, an intersection type would resolve to never because for example, there is no intersection between `{ status: 200; body: any } & { status: 400; body unknown }`, so you end up with `never`

```suggestion
export type ApiResponseForRoute<T extends AppRoute> =
  | {
      [K in keyof T['responses']]: {
        status: K
        body: ZodInferOrType<T['responses'][K]>
      }
    }[keyof T['responses']]
  | {
      status: Exclude<HTTPStatusCode, keyof T['responses']>
      body: unknown
    };
```

arivera-xealth: AH, you're totally right! Thank you! This is perfect 👏 a silly mistake on my end.

arivera-xealth: So that's why `ApiRouteResponse<T['responses']>` should just work, since it's essentially the same and you're passing in that type

michaelangrivera: AH, you're totally right! Thank you! This is perfect 👏 a silly mistake on my end.

michaelangrivera: So that's why ApiRouteResponse<T['responses']> should just work, since it's essentially the same and you're passing in that type

libs/ts-rest/core/src/lib/client.ts
oliverbutler: I never had time to tackle this, but I don't want to block this release, I'll open an issue :)

PR #115, opened 2022-12-29
Optional JSON-encoded query parameters

Gabrola:
Now the library can truly claim to offer "Fully typed RPC-like client"

I think the main point of discussion would be the opinionated approach of the selective encoding format of root-level strings.

The implementation, as is, will not JSON encode (i.e. surround with double quotes) a root-level string unless it clashes with a JSON reserved word or is a number.

```typescript
const query = { foo: 'foo', nested: { bar: 'bar' }  };
// ?foo=foo&nested={"bar":"bar"}

const query = { foo: 'true' };
// ?foo="true"

const query = { foo: true };
// ?foo=true

const query = { foo: '123' };
// ?foo="123"

const query = { foo: 123 };
// ?foo=123
```

I think it's nice because this leaves the URL looking like regular query strings in the majority of use cases. This is also how swagger encodes root-level strings even if we specify the MIME type as `application/json`

Screenshot 2022-12-29 062946

### DIFF
--- a/libs/ts-rest/core/src/lib/query.ts
+++ b/libs/ts-rest/core/src/lib/query.ts
@@ -1,11 +1,41 @@
 /**
  *
  * @param query - Any JSON object
+ * @param json - Use JSON.stringify to encode the query values
  * @returns - The query url segment, using explode array syntax, and deep object syntax
  */
-export const convertQueryParamsToUrlString = (query: unknown) => {
-  const queryString = encodeQueryParams(query);
+export const convertQueryParamsToUrlString = (query: unknown, json = false) => {
+  const queryString = json
+    ? encodeQueryParamsJson(query)
+    : encodeQueryParams(query);
   return queryString?.length > 0 ? '?' + queryString : '';
+};
+
+export const encodeQueryParamsJson = (query: unknown) => {
+  if (!query) {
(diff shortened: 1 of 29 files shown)


arivera-xealth:
this is sick!

oliverbutler:
I really like this approach! I'm going to review now, however I notice there is a merge conflict @Gabrola - If it's simple I'll try fix it now.

libs/ts-rest/core/src/lib/query.ts
```diff
@@ -1,13 +1,43 @@
...
-export const convertQueryParamsToUrlString = (query: unknown) => {
-  const queryString = encodeQueryParams(query);
+export const convertQueryParamsToUrlString = (query: unknown, json = false) => {
+  const queryString = json
+    ? encodeQueryParamsJson(query)
+    : encodeQueryParams(query);
   return queryString?.length > 0 ? '?' + queryString : '';
 };
 
+export const encodeQueryParamsJson = (query: unknown) => {
```
oliverbutler, line 14: Wonderfully simple implementation, I love it ❤️

oliverbutler (approved):

Gabrola:
Cheers mate! ♥

PR #116, opened 2022-12-30
Fix typing for query

Gabrola:
Looks like a small issue was introduced with the `AreAllPropertiesOptional` change. `query` was resolving to `never` if not all of its options were optional, but also for some reason the typing was still broken even they were all optional.

Screenshot 2022-12-30 155037

You can see here, typescript was not complaining at all, despite `null` not being an compatible value.


oliverbutler:
Hey @Gabrola! Thanks for fixing this, I'm back from my holiday leave now so happy to merge.

PR #121, opened 2023-01-13
feat: add response validation functionality to express, next and nest

lukemorales:
This PR injects the `appRoute` to the `ApiRouteInterceptor` so that zod validations and transformations can run on the response body


Gabrola:
This is great! I believe this is something a lot of people might want.

One thing to note, not everyone would probably want to validate their responses, and would be satisfied with just the type-safety. So I'd consider it essential to make sure this is optional and disabled by default otherwise this would be a breaking change.

libs/ts-rest/nest/src/lib/api.decorator.ts
```diff
@@ -141,6 +162,6 @@ export const Api = (appRoute: AppRoute): MethodDecorator => {
   return applyDecorators(
     SetMetadata(tsRestAppRouteMetadataKey, appRoute),
     methodDecorator,
-    UseInterceptors(ApiRouteInterceptor)
+    UseInterceptors(new ApiRouteInterceptor(appRoute))
```
Gabrola, line 165: This will create a new `ApiRouteInterceptor` instance for each endpoint, therefore increasing the memory usage of the process significantly. As the `appRoute` is already part of the request handler's metadata you can access it in the interceptor using this code on lines 52-55

```typescript
    const appRoute: AppRoute | undefined = Reflect.getMetadata(
      tsRestAppRouteMetadataKey,
      context.getHandler()
    );
```

libs/ts-rest/nest/src/lib/api.decorator.ts
```diff
@@ -115,18 +116,38 @@ const getMethodDecorator = (appRoute: AppRoute) => {
...
+        if (this.isAppRouteResponse(value)) {
+          const statusNumber = value.status;
+
+          const responseValidation = checkZodSchema(
+            value.body,
+            this.appRoute.responses[statusNumber]
+          );
+
+          if (!responseValidation.success) {
+            throw new BadRequestException(responseValidation.error);
```
Gabrola, line 146: I'd imagine a response validation error would qualify as a 5xx error than than a 400 error

libs/ts-rest/nest/src/lib/api.decorator.ts
```diff
@@ -115,18 +116,38 @@ const getMethodDecorator = (appRoute: AppRoute) => {
 
 @Injectable()
 export class ApiRouteInterceptor implements NestInterceptor {
+  constructor(private readonly appRoute: AppRoute) {}
```
michaelangrivera, line 119: Can we make this constructor accept a param to mark this feature as optional? Similar to what @Gabrola suggested.

libs/ts-rest/nest/src/lib/api.decorator.ts
michaelangrivera: Can we make that response error configurable? I can see cases where you'd want to customize the error based on the zod validation.

michaelangrivera:
> This is great! I believe this is something a lot of people might want.
> 
> One thing to note, not everyone would probably want to validate their responses, and would be satisfied with just the type-safety. So I'd consider it essential to make sure this is optional and disabled by default otherwise this would be a breaking change.

Agree. Also in case @oliverbutler marks the `zod` peer dependency as optional in future versions, we should absolutely mark this as an opt in feature.

michaelangrivera (changes requested):
Please make this feature an opt-in! Great work 🍾

oliverbutler:
@michaelangrivera Zod IS optional, in that if you don't use it, zod won't get bundled into your contract

but there's an outstanding issue to remove references to zod from the library so zod doesn't need to be a dev dependency 

this looks amazing! i've been tight on time this last week but i should be able to take a look today 🤞

thanks for the PR!

libs/ts-rest/nest/src/lib/api.decorator.ts
Gabrola: You  could do that already using [exception filters](https://docs.nestjs.com/exception-filters#exception-filters-1) on HttpExceptions and check if `res.body.name === 'ZodError'`, then rewrite the response status and body if needed, but it would be cleaner if we use a custom error type and filter, but I guess that's a discussion for another thread.

libs/ts-rest/nest/src/lib/api.decorator.ts
Gabrola: We shouldn't pass values manually to injectable class constructors because the DI container would fail to instantiate new instances if needed. The NestJS way would be to create a custom decorator for example `@ValidateResponses()`, and use metadata to check if it's enabled.

libs/ts-rest/nest/src/lib/api.decorator.ts
lukemorales: Nice, thanks for pointing that out @Gabrola

lukemorales:
Just pushed the requested changes and made the parsing opt-in.

The way I achieved this was at the controller level by using the new `@ParseResponses` decorator, and by introducing a new parameter to `initNestServer` that will change the responses types from `z.infer` to `z.input` because the response will need to match the input of zod instead of the output

```ts
const s = initNestServer(apiBlog, { parseResponses: true });

@Controller()
@ParseResponses()
export class PostController implements ControllerShape {
  constructor(private readonly postService: PostService) {}

  @Api(s.route.getPost)
  async getPost(@ApiDecorator() { params: { id } }: RouteShape['getPost']) {
    const post = await this.postService.getPost(id);

    if (!post) {
      return { status: 404 as const, body: null };
    }

    return { status: 200 as const, body: post };
  }
}
```

libs/ts-rest/nest/src/lib/api.decorator.ts
```diff
@@ -115,18 +118,57 @@ const getMethodDecorator = (appRoute: AppRoute) => {
...
+            value.body,
+            isParsingEnabled ? appRoute.responses[statusNumber] : undefined
+          );
+
+          if (!responseValidation.success) {
+            const { error } = responseValidation;
+
+            const message = error.errors.map(
+              (error) => `${error.path.join('.')}: ${error.message}`
+            );
```
lukemorales, line 165: Using the same approach as `@anatine/zod-nestjs` to better formatted validation errors

libs/ts-rest/nest/src/lib/ts-rest-nest.ts
```diff
@@ -21,18 +34,45 @@ type AppRouteShape<T extends AppRouter> = {
...
+
+export const initNestServer = <
+  T extends AppRouter,
+  Options extends InitServerOptions
+>(
+  router: T,
+  options?: Options
+): Options['parseResponses'] extends true
+  ? InitServerReturn<T, true>
+  : InitServerReturn<T, false> => {
```
lukemorales, line 66: Another approach can be adding another generic:
```
export const initNestServer = <
  T extends AppRouter,
  Options extends InitServerOptions,
  ParseResponses extends Options['parseResponses'] = false
>(
  router: T,
  options?: Options
): InitServerReturn<T, ParseResponses>
```

libs/ts-rest/nest/src/lib/api.decorator.ts
oliverbutler, line 165: Does this succeed if isParsingEnabled is false, leading the second param of checkZodSchema to be undefined? 

Alternatively, maybe we should early return if the validation isn't on 🤔 

Also, I'm aware this is a breaking change, primarily if someone is relying on receiving the Zod errors as an object.  On the flip side, the Zod error handling was undocumented behaviour, maybe we should discuss this change, then document + test it.

libs/ts-rest/nest/src/lib/api.decorator.ts
```diff
@@ -115,18 +118,57 @@ const getMethodDecorator = (appRoute: AppRoute) => {
...
 
+    const appRoute: AppRoute | undefined = Reflect.getMetadata(
+      tsRestAppRouteMetadataKey,
+      context.getHandler()
+    );
+
+    if (!appRoute) {
+      // this will respond with a 500 error without revealing this error message in the response body
+      throw new Error('Make sure your route is decorated with @Api()');
+    }
```
michaelangrivera, line 143: Great addition 🥳

libs/ts-rest/nest/src/lib/api.decorator.ts
```diff
@@ -115,18 +118,57 @@ const getMethodDecorator = (appRoute: AppRoute) => {
...
+
     return next.handle().pipe(
       map((value) => {
```
michaelangrivera, line 146: this has `any` type, implicitly. Can we type this at all?

michaelangrivera: Nevermind, I see `isAppRouteResponse` provides the typing. Thanks!

michaelangrivera:
> Does this succeed if isParsingEnabled is false, leading the second param of checkZodSchema to be undefined?
> 
> Alternatively, maybe we should early return if the validation isn't on 🤔
> 
> Also, I'm aware this is a breaking change, primarily if someone is relying on receiving the Zod errors as an object. On the flip side, the Zod error handling was undocumented behaviour, maybe we should discuss this change, then document + test it.

Agree. We should definitely exit early if that decorator does not exist. Something like: 
```
          const statusNumber = value.status;

          const isParsingEnabled = Boolean(
            Reflect.getMetadata(ParseResponsesSymbol, context.getHandler()) ||
              Reflect.getMetadata(ParseResponsesSymbol, context.getClass())
          );
          
          if (!isParsingEnabled) {
            res.status(statusNumber)
            return value.body
          } 
```
would read easier in my opinion. 

Sorry to cause extra work here, but is it possible to separate this logic into another method or util function? You might be able to move to it `parse-responses.decorator.ts`.
```
          const responseValidation = checkZodSchema(
            value.body,
            isParsingEnabled ? appRoute.responses[statusNumber] : undefined
          );

          if (!responseValidation.success) {
            const { error } = responseValidation;

            const message = error.errors.map(
              (error) => `${error.path.join('.')}: ${error.message}`
            );

            throw new InternalServerErrorException(message);
          }
```
Should we migrate to zod being an optional devDependency, this would make that migration easier. We should start doing that from here on out (imo). 

@oliverbutler  If we return early, I'm not sure this would be a breaking change?  The logic only runs if the `ParseResponses` decorator exists from the looks of it. Can you point to the zod errors aspect for response parsing (on `main` branch)? 

## A few other notes

- Can you write a few tests demonstrating this? You should be able to use `post.controller.spec.ts` and `post.controller.withResponseShapes.spec.ts` for an "integration" style test.  
- Also, can we unit test:
```
    const responseValidation = checkZodSchema(
           ....
          ⬇️ 
          ...
            throw new InternalServerErrorException(message);
          }
``` 
- That way, if `zod` changes anything (unlikely), we'll know. 
- Can you create an example file off of `post.controller.ts` that shows this feature working? Or even modify the existing one with a new method showing this new decorator in action. 
- Can you update the `nest` documentation with any and all new information about this decorator? You could add a new section on "Response Parsing" or something similar. Please give the nest docs a read if you haven't ready to see the style of how we document features: https://ts-rest.com/docs/nest


Great work here! Thank you so much! Will definitely be useful.

libs/ts-rest/nest/src/lib/api.decorator.ts
Gabrola: @oliverbutler indeed that would be a breaking change, we consume the raw Zod errors in an exception filter, and rewrite them into our own error format.

@lukemorales can we separate this change into a separate issue to be discussed further?

lukemorales: Sure thing, no problem! I thought of suggesting this approach for this as this would be a case were errors weren't being thrown for validation at the first place and perhaps a better formatted error message would be helpful to debug what's wrong with the parsed data. But it makes a lot of sense that people might be catching these errors and I will revert back to the original zod error

lukemorales: >Does this succeed if isParsingEnabled is false, leading the second param of checkZodSchema to be undefined?
>
> Alternatively, maybe we should early return if the validation isn't on 🤔

It does success according to the `checkZodSchema` implementation, because the type guard accounts for undefined in it's implementation, there are even test cases written for that: https://github.com/ts-rest/ts-rest/blob/cb74bab9e26a8254adf4a05dc5e78c18b2249ff2/libs/ts-rest/core/src/lib/zod-utils.spec.ts#L41-L52

But just sharing for visibility, I agree with you and @michaelangrivera's comment that early return would be better for visibility

oliverbutler (approved):
Looks great!

michaelangrivera:
LGTM!

michaelangrivera (approved):

apps/docs/docs/nest.md
```diff
@@ -37,38 +37,56 @@ To implement a nested contract, you can simply call `nestControllerContract(cont
...
+This is useful for ensuring absolute safety that your controller is returning the correct response types as well as stripping any extra properties.
+
+```typescript
+@Controller()
+@TsRest({ validateResponses: true })
+export class PostController implements NestControllerInterface<typeof c> {}
+```
+
+If validation fails a `ResponseValidationError` will be thrown causing a 500 response to be returned.
+You can catch this error and handle it as you wish by using a [NestJS exception filter](https://docs.nestjs.com/exception-filters).
```
michaelangrivera, line 88: this is great!

Gabrola:
@michaelangrivera @oliverbutler cheers!

@lukemorales thank you so much for taking on this initiative. Very excited to see this go live! 🙌

lukemorales:
Thanks for pushing it to the finish line when I had trouble finding time to come back to it @Gabrola, really appreciate the support!

PR #128, opened 2023-01-20
Add useQueries to @tanstack/react-query

### LINKED ISSUE #127: Support useQueries in @ts-rest/react-query
I just had a chat with a colleague who needed to use useQueries with react query.

I realised we don't support this! We support useQuery and useInfiniteQuery

Maybe we could use a proposed API like this? It closely reflects the RQ api.
image
image

oliverbutler:
Fixes #127

### DIFF
--- a/.changeset/spotty-cherries-camp.md
+++ b/.changeset/spotty-cherries-camp.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/react-query': minor
+---
+
+Add useQueries capability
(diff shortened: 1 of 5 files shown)


oliverbutler:
This is missing the ability for the typed response to be a tuple, rather than an unbounded array, means users will need to do an is not undefined check

oliverbutler:
Also missing documentation

Gabrola:
LGTM!

PR #140, opened 2023-01-23
Revised Nest function and type names

### LINKED ISSUE #122: Passing Error object directly to HttpException is deprecated (NestJS)
Passing the Zod validation error directly to the `BadRequestException` constructor is now triggering this warning

`DEPRECATED! Passing the error cause as the first argument to HttpException constructor is deprecated. You should use the "options" parameter instead: new HttpException("message", 400, { cause: new Error("Some Error") })`

Gabrola:
I've renamed some of the Nest function and type names as they have confused some people.

I've also changed how the types are consumed rather than using the `{} as T` hack. In my opinion, this makes it clearer that we are providing types rather than running any significant code at runtime.

Also updated the docs to better explain the role of every function and decorator, and how the type safety is done.
 
Also fixes #122

### DIFF
--- a/.changeset/dull-maps-draw.md
+++ b/.changeset/dull-maps-draw.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/nest': minor
+---
+
+Rename some Nest functions and types, and deprecate old names
--- a/.changeset/dull-maps-draw.md
+++ b/.changeset/dull-maps-draw.md
@@ -6,3 +6,6 @@
 Rename some Nest functions and types, and deprecate old names
 
 Fix Nest deprecation warning when passing Zod error to HttpException (#122)
+
+Some internal helper types (`NestControllerShapeFromAppRouter` and `NestAppRouteShape`) that were previously exported are now kept internal.
+You can use `NestControllerInterface` and `NestRequestShapes` instead.
--- a/.changeset/dull-maps-draw.md
+++ b/.changeset/dull-maps-draw.md
@@ -1,5 +1,8 @@
 ---
 '@ts-rest/nest': minor
+'@ts-rest/core': patch
 ---
 
 Rename some Nest functions and types, and deprecate old names
+
+Fix Nest deprecation warning when passing Zod error to HttpException (#122)
--- a/apps/docs/docs/core/form-data.mdx
+++ b/apps/docs/docs/core/form-data.mdx
@@ -116,7 +116,7 @@ export class AppController implements NestControllerInterface<typeof c> {
   @Api(s.route.updateUserAvatar)
   @UseInterceptors(FileInterceptor('avatar'))
   async updateUserAvatar(
-    @TypedRequest() { params: { id } }: RequestShapes['updateUserAvatar'],
+    @TsRestRequest() { params: { id } }: RequestShapes['updateUserAvatar'],
     @UploadedFile() avatar: Express.Multer.File
   ) {
     return {
--- a/apps/docs/docs/core/form-data.mdx
+++ b/apps/docs/docs/core/form-data.mdx
@@ -112,11 +112,11 @@ With Nest this is a pretty simple implementation, due to the extensible Decorato
 ```ts
 // nest
 @Controller()
-export class AppController implements ControllerShape {
+export class AppController implements NestControllerInterface<typeof c> {
   @Api(s.route.updateUserAvatar)
   @UseInterceptors(FileInterceptor('avatar'))
   async updateUserAvatar(
-    @ApiDecorator() { params: { id } }: RouteShape['updateUserAvatar'],
+    @TypedRequest() { params: { id } }: RequestShapes['updateUserAvatar'],
     @UploadedFile() avatar: Express.Multer.File
   ) {
     return {
--- a/apps/docs/docs/nest.md
+++ b/apps/docs/docs/nest.md
@@ -1,19 +1,25 @@
 # Nest
 
-By default Nest doesn't offer a nice way to ensure type safe controllers, primarily because it's decorator driven rather than functional, like Express.
+As opposed to the Express implementation, where we can infer types from a function signature, we need to explicitly define our types since Nest controllers are implemented as classes.
 
 ```typescript
-const s = initNestServer(apiBlog);
-type ControllerShape = typeof s.controllerShape;
-type RouteShape = typeof s.routeShapes;
-type ResponseShapes = typeof s.responseShapes;
+import {
+  Api,
+  nestControllerContract,
+  NestControllerInterface,
+  NestRequestShapes,
+  TypedRequest,
+} from '@ts-rest/nest';
+
+const c = nestControllerContract(apiBlog);
+type RequestShapes = NestRequestShapes<typeof c>;
 
 @Controller()
-export class PostController implements ControllerShape {
+export class PostController implements NestControllerInterface<typeof c> {
   constructor(private readonly postService: PostService) {}
 
-  @Api(s.route.getPost)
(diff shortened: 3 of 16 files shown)


libs/ts-rest/nest/src/lib/ts-rest-nest.ts
```diff
@@ -1,33 +1,48 @@
...
+type NestControllerShapeFromAppRouter<T extends AppRouter> = Without<
   AppRouterControllerShape<T>,
   AppRouter
 >;
 
-export type NestAppRouteShape<T extends AppRouter> = AppRouteShape<T>;
+type NestAppRouteShape<T extends AppRouter> = AppRouterRequestShapes<T>;
 
+/**
+ * @deprecated Use `nestControllerContract`, `NestControllerInterface`, `NestRequestShapes`, and `NestResponseShapes` instead. Check the docs for more info.
```
oliverbutler, line 44: Can we link to the nest docs to save users some time?

oliverbutler:
I want to pull this and play with it locally, but this looks great! 

I wonder if this is an opportunity to discuss adding codemods for migrations in ts-rest? I've never done them myself so not sure on the effort required on our side to make them.

libs/ts-rest/nest/src/lib/ts-rest-nest.ts
Gabrola: Absolutely!

Gabrola:
> I want to pull this and play with it locally, but this looks great!
> 
> I wonder if this is an opportunity to discuss adding codemods for migrations in ts-rest? I've never done them myself so not sure on the effort required on our side to make them.

I've thought about it too. I think it could be a good idea to do during the next major release if we decide to drop the deprecated versions.

.changeset/dull-maps-draw.md
```diff
@@ -0,0 +1,8 @@
+---
+'@ts-rest/nest': minor
+'@ts-rest/core': patch
+---
+
+Rename some Nest functions and types, and deprecate old names
```
michaelangrivera, line 6: Since we're renaming, should we indicate breaking changes in the release notes?

apps/docs/docs/core/form-data.mdx
```diff
@@ -112,11 +112,11 @@ With Nest this is a pretty simple implementation, due to the extensible Decorato
 ```ts
 // nest
 @Controller()
-export class AppController implements ControllerShape {
+export class AppController implements NestControllerInterface<typeof c> {
   @Api(s.route.updateUserAvatar)
   @UseInterceptors(FileInterceptor('avatar'))
   async updateUserAvatar(
-    @ApiDecorator() { params: { id } }: RouteShape['updateUserAvatar'],
+    @TypedRequest() { params: { id } }: RequestShapes['updateUserAvatar'],
```
michaelangrivera, line 119: I'm wondering if `TypedRequest` should just be named something to reflect the package name itself? 

For instance, importing `{ z } from 'zod'` feels super natural. So maybe something like, `TSRest` or `TSRestRoute` would look even better and allow people to easily distinguish when they're using the `ts-rest` library?

michaelangrivera: So we would have:

```
import { TSRestRequest } from 'ts-rest/nest'

 async updateUserAvatar(
    @TSRestRequest() { params: { id } }: TSRestShapes['updateUserAvatar'],
```

apps/docs/docs/nest.md
```diff
@@ -36,63 +50,61 @@ To handle JSON query parameters, you can use the `@JsonQuery()` decorator on eit
...
 
-## Response Return Type Safety
-
-You have two options to ensure HTTP type safety on your Nest Controllers:
-
-- `ControllerShape` as shown above:
-  - Your controller can implement the `ControllerShape` derived from `typeof s.controllerShape`
-  - This ensures your controller methods also align with the base interface shapes
-- `ResponseShapes`:
+## Explicit Response Type Safety
```
michaelangrivera, line 72: Love this.

libs/ts-rest/nest/src/lib/ts-rest-nest.ts
```diff
@@ -1,33 +1,49 @@
...
+  never
+>;
 
-export type NestControllerShapeFromAppRouter<T extends AppRouter> = Without<
+type NestControllerShapeFromAppRouter<T extends AppRouter> = Without<
   AppRouterControllerShape<T>,
   AppRouter
 >;
 
-export type NestAppRouteShape<T extends AppRouter> = AppRouteShape<T>;
```
michaelangrivera, line 29: We should communicate we are removing this from exports. If people are using this (somehow), that's a breaking change.

libs/ts-rest/nest/src/lib/ts-rest.interceptor.ts
```diff
@@ -0,0 +1,65 @@
...
+  SetMetadata,
+  UseInterceptors,
+} from '@nestjs/common';
+import { map, Observable } from 'rxjs';
+import { Response } from 'express-serve-static-core';
+import { AppRoute } from '@ts-rest/core';
+import { tsRestAppRouteMetadataKey } from './typed-request.decorator';
+
+@Injectable()
+export class TsRestInterceptor implements NestInterceptor {
```
michaelangrivera, line 21: Exactly what I was saying, `TsRestInterceptor` makes me _know_ that I'm using ts-rest. I love this personally.

.changeset/dull-maps-draw.md
Gabrola: We're still keeping the old names, but marking them as deprecated so it's technically not a breaking change yet

michaelangrivera: AH missed that, thanks!

apps/docs/docs/core/form-data.mdx
Gabrola: I agree fully. That was actually the initial name I came up with, but I felt like `TSRestRequest` was a little clunky to type and a bit of tongue-twister 😅

Let's think of something

libs/ts-rest/nest/src/lib/ts-rest-nest.ts
Gabrola: I believe this was an unintended export, as it was never meant to be used directly. I'm going back to @oliverbutler's comment on avoiding breaking changes the second we document our API (which we did but not too long ago, I doubt anyone picked this one up specifically from the docs yet 😅).

I'm personally fine with the trade-off of potentially breaking some code using undocumented features for the purpose of keeping our code more maintainable. Especially since this specific example would be very easy to fix.

I guess, rather than bumping the major version for such a small change, we could at least add it to the release notes, in case this affects some people.

michaelangrivera: Yup, that was my point, let's at a minimum, add it to the release notes.

.changeset/dull-maps-draw.md
```diff
@@ -0,0 +1,8 @@
...
+
+Fix Nest deprecation warning when passing Zod error to HttpException (#122)
```
michaelangrivera, line 8: Should we add the removal of the exports here? 

Or does that go somewhere else?

michaelangrivera (approved):
LGTM! Will re-approve should you add anything to changeset

.changeset/dull-maps-draw.md
Gabrola: Yup, here sounds about right in this case

oliverbutler (approved):

oliverbutler:
This looks great! LGTM :)

oliverbutler:
I way prefer the naming, and awesome that you polled the discord 🔥

PR #142, opened 2023-01-24
Drop usage of ES6 Proxy

### LINKED ISSUE #131: IE11 Support
I'm getting `'Proxy' is undefined` in IE11.

Would Polyfiling Proxy be the only obstacle to IE11 usage? Is there any plan to support IE11?

At my company, we've switched from tRPC to ts-rest in the hope that we can use it in IE11.

Gabrola:
Fixes #131

### DIFF
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -70,8 +70,9 @@ export type ApiRouteResponse<T> =
       body: unknown;
     };
 
-export type ApiResponseForRoute<T extends AppRoute> = ApiRouteResponse<T['responses']>
-
+export type ApiResponseForRoute<T extends AppRoute> = ApiRouteResponse<
+  T['responses']
+>;
 
 /**
  * Returned from a mutation or query call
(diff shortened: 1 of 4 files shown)


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -202,45 +203,30 @@ export const getRouteQuery = <TAppRoute extends AppRoute>(
...
 ): InitClientReturn<T> => {
-  const proxy = createNewProxy(router, args);
-
-  return proxy as InitClientReturn<T>;
+  return Object.fromEntries(
+    Object.entries(router).map(([key, subRouter]) => {
+      if (isAppRoute(subRouter)) {
+        return [key, getRouteQuery(subRouter, args)];
+      } else {
+        return [key, initClient(subRouter, args)];
```
stingrayfunk, line 217: Really minor thing, but you don't need to wrap an else around this return.

libs/ts-rest/react-query/src/lib/react-query.ts
```diff
@@ -340,45 +340,29 @@ const getRouteUseMutation = <TAppRoute extends AppRoute>(
...
+            query: getRouteQuery(subRouter, args),
+            mutation: getRouteQuery(subRouter, args),
+            useQuery: getRouteUseQuery(subRouter, args),
+            useQueries: getRouteUseQueries(subRouter, args),
+            useInfiniteQuery: getRouteUseInfiniteQuery(subRouter, args),
+            useMutation: getRouteUseMutation(subRouter, args),
+          },
+        ];
+      } else {
+        return [key, initQueryClient(subRouter, args)];
```
stingrayfunk, line 364: Return does not need to be wrapped in an else.

libs/ts-rest/solid-query/src/lib/solid-query.ts
```diff
@@ -246,43 +246,28 @@ const getRouteUseMutation = <TAppRoute extends AppRoute>(
...
+          {
+            query: getRouteQuery(subRouter, args),
+            mutation: getRouteQuery(subRouter, args),
+            createQuery: getRouteUseQuery(subRouter, args),
+            createInfiniteQuery: getRouteUseInfiniteQuery(subRouter, args),
+            createMutation: getRouteUseMutation(subRouter, args),
+          },
+        ];
+      } else {
+        return [key, initQueryClient(subRouter, args)];
```
stingrayfunk, line 269: Return does not need to be wrapped in an else.

libs/ts-rest/core/src/lib/client.ts
Gabrola: True, it's technically redundant, but it depends on the context and how it improves readability. Dropping the else makes sense when you want to return early or are using a guard pattern. Semantically, in this case, the if/else helps readability. It's a style preference at the end of the day.

michaelangrivera: Personally, I like exit early patterns, but this looks fine in this case.

oliverbutler:
@Gabrola I like it! Going to put some dinner on and have a read though but a cursory glance looks great :D

oliverbutler (approved):

PR #145, opened 2023-01-24
fix: clone the response body before reading json

### LINKED ISSUE #141: TypeError: Failed to execute 'text' on 'Response': body stream already read
Hi, first of all, congrats making this really good library !

I'm using `ts-rest` inside an NX Monorepo with `NextJS` & `NestJS`.

I got my two apps (generated form `@nrwl/nest` & `@nrwl/next`) and a lib for the contracts.

Here are my definitions :

```ts
// index.ts in shared/contracts/src (my library)

const c = initContract();

export const healthCheckContract = c.router({
  getHealthcheck: {
    method: 'GET',
    path: '/',
    responses: {
      200: c.response<string>(),
    },
    summary: 'Healthcheck API to check if the service is up and running',
  },
});
```

```ts
// app.controller.ts in NestJS app
import { Controller } from '@nestjs/common';
import {Api, initNestServer} from "@ts-rest/nest";
import {healthCheckContract} from "@ecommerce/contracts";

const s = initNestServer(healthCheckContract);
type ControllerShape = typeof s.controllerShape;
type ResponseShapes = typeof s.responseShapes;

@Controller()
export class AppController implements ControllerShape {
  @Api(s.route.getHealthcheck)
  async getHealthcheck(): Promise<ResponseShapes['getHealthcheck']> {
    return  {
      status: 200,
      body: 'Hello World!'
    };
  }
}
```

```ts
// _app.tsx in NextJS app (init queryClientProvider)
import { AppProps } from 'next/app';
import Head from 'next/head';
import {QueryClient, QueryClientProvider} from "@tanstack/react-query";
import {useState} from "react";

export default function CustomApp({ Component, pageProps }: AppProps) {
  const [queryClient] = useState(() => new QueryClient());

  return (
    <>
      <Head>
        <title>Welcome to web!</title>
      </Head>
      <main className="app">
        <QueryClientProvider client={queryClient}>
            <Component {...pageProps} />
        </QueryClientProvider>
      </main>
    </>
  );
}
```

```ts
// index.tsx in NextJS app (fetching data)
import {initQueryClient} from "@ts-rest/react-query";
import {healthCheckContract} from "@test/contracts";
import {useEffect} from "react";

export default function Index() {
  const client = initQueryClient(healthCheckContract, {
    baseUrl: 'http://localhost:3333/api',
    baseHeaders: {},
  });
  const {data} = client.getHealthcheck.useQuery(['healthcheck'])

  useEffect(() => {
    console.log('data', data)
  }, [data])

  return (
    <div>
      <h1>Welcome to web!</h1>
    </div>
  );
}
```

I can watch the query response, and from there, all good : 

image

The console.log pops one time, and says "undefined"
Then this error appears in the console :

image

Coming from this part of code : 

image

And, as I can see, in network tab, the query repeats itself again and again.

No type errors so I presume my code is correct but maybe there is something I miss ?
Thank you for your help.

oliverbutler:
Fixes #141 

@Gabrola @michaelangrivera I think this fixes the issue, cloning the body before we try to convert it to json.

### DIFF
--- a/libs/ts-rest/core/src/lib/client.spec.ts
+++ b/libs/ts-rest/core/src/lib/client.spec.ts
@@ -293,6 +293,23 @@ describe('client', () => {
         body: undefined,
       });
     });
+
+    it('w/ a non json response (string)', async () => {
+      api.mockResolvedValue({ body: 'string', status: 200 });
+
+      const result = await client.posts.getPosts({});
+
+      expect(result).toStrictEqual({ body: 'string', status: 200 });
+
+      expect(api).toHaveBeenCalledWith({
+        method: 'GET',
+        path: 'https://api.com/posts',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: undefined,
+      });
+    });
   });
 
   describe('post', () => {
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -114,7 +114,7 @@ export const defaultApi: ApiFetcher = async ({
   try {
     return {
       status: result.status,
-      body: await result.json(),
+      body: await result.clone().json(),
     };
   } catch {
     return {
--- a/apps/example-next/tests/react-query.spec.tsx
+++ b/apps/example-next/tests/react-query.spec.tsx
@@ -182,6 +182,38 @@ describe('react-query', () => {
     expect(result.current.data).toStrictEqual(SUCCESS_RESPONSE);
   });
 
+  it('useQuery should accept non-json string response', () => {
+    api.mockResolvedValue({
+      status: 200,
+      body: 'Hello World',
+    });
+
+    const { result } = renderHook(() => client.health.useQuery(['health']), {
+      wrapper,
+    });
+
+    expect(result.current.data).toStrictEqual(undefined);
+
+    expect(result.current.isLoading).toStrictEqual(true);
+
+    expect(api).toHaveBeenCalledWith({
+      method: 'GET',
+      path: 'http://api.com/health',
+      body: undefined,
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+
+    return waitFor(() => {
+      expect(result.current.isLoading).toStrictEqual(false);
+      expect(result.current.data).toStrictEqual({
+        status: 200,
+        body: 'Hello World',
+      });
+    });
+  });
+
   it('useQuery should handle failure', async () => {
     api.mockResolvedValue(ERROR_RESPONSE);
 


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -114,7 +114,7 @@ export const defaultApi: ApiFetcher = async ({
   try {
     return {
       status: result.status,
-      body: await result.json(),
+      body: await result.clone().json(),
```
michaelangrivera, line 117: any memory implications with this? 🤔

Gabrola: Yeah, based on [this](https://developer.mozilla.org/en-US/docs/Web/API/Response/clone)

> If only one cloned branch is consumed, then the entire body will be buffered in memory.

But, I'm pretty sure the GC will take care of it.

libs/ts-rest/core/src/lib/client.spec.ts
```diff
@@ -293,6 +293,23 @@ describe('client', () => {
         body: undefined,
       });
     });
+
+    it('w/ a non json response (string)', async () => {
```
Gabrola, line 297: It's nice to have these tests anyway, but they wouldn't test this specific problem since we're mocking the api entirely, so no fetch is called or anything.

libs/ts-rest/core/src/lib/client.ts
Gabrola: Or to just get rid of any doubt maybe we can read it once as text, then `try { return JSON.parse(text); } catch { return text }`

michaelangrivera:
Perhaps this could be brought up as another issue, however, based on the following examples: 
https://stackoverflow.com/questions/47989559/javascript-how-to-know-the-type-of-the-response-in-fetch-api 
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#headers

And specifically: 
```
When extract a MIME type returns failure or a MIME type whose essence is incorrect for a given format, treat this as a fatal error. Existing web platform features have not always followed this pattern, which has been a major source of security vulnerabilities in those features over the years. In contrast, a MIME type’s parameters can typically be safely ignored. https://fetch.spec.whatwg.org/#content-type-header
```
and from MDN
```
A good use case for headers is checking whether the content type is correct before you process it further. For example:

fetch(myRequest)
  .then((response) => {
     const contentType = response.headers.get('content-type');
     if (!contentType || !contentType.includes('application/json')) {
       throw new TypeError("Oops, we haven't got JSON!");
     }
     return response.json();
  })
  .then((data) => {
      /* process your data further */
  })
  .catch((error) => console.error(error));
```

Maybe we _should_ utilize the `content-type` header and fail accordingly if it's not present. That way, we're implementing logic that includes potentially expensive calls, like the try-catch block:
```
  try {
    return {
      status: result.status,
      body: await result.clone().json(),
    };
  } catch {
```
And on top of that, we're following the security practices mentioned above. 

Taking the above into account, you could write something like this:
```
  const result = await fetch(path, { method, headers, body, credentials });
  if (!result.headers.get('content-type'))
    throw TypeError('Resource Response missing content-type header')
  const blob = await result.blob()
  if (blob.type === 'application/json') {
    return { status: result.status, body: await result.json() }
  }
  if (blob.type === 'text/plain') {
    return { status: result.status, body: await result.text() }
  }
  return { status: result.status, body: blob }
```
And, mention in the docs that we support basic fetch calls for `application/json` and `text/plain` content types and do not support `content-type` missing in resource response. If they need to implement custom API logic (like content-type header is missing for whatever reason), point them to the custom API docs.

libs/ts-rest/core/src/lib/client.spec.ts
```diff
@@ -293,6 +293,23 @@ describe('client', () => {
...
+
+      const result = await client.posts.getPosts({});
+
+      expect(result).toStrictEqual({ body: 'string', status: 200 });
+
+      expect(api).toHaveBeenCalledWith({
+        method: 'GET',
+        path: 'https://api.com/posts',
+        headers: {
+          'Content-Type': 'application/json',
```
michaelangrivera, line 308: the header should be `text/plain` in this case, I believe

libs/ts-rest/core/src/lib/client.ts
michaelangrivera: Yeah, not sure if we should rely on the already expensive GC that is Javascript/Node LOL.

PR #151, opened 2023-01-29
Fix incorrect @ts-react/react-query package name in docs

rifaldhiaw:
Hi, thanks for the awesome library!

I just tried it out with next.js as a backend implementation. While I was reading the doc, I found that there is a wrong package name in the documentation. I realized it when I try to install it and it is not found. Turns out the right package name was the one in the sidebar. So, I want to contribute a little to fix the doc.

Thanks


Gabrola:
Awesome, thanks for your contribution!

Gabrola (approved):
Awesome, thanks for your contribution!

Gabrola (approved):

PR #157, opened 2023-02-02
refactor: type safe TsRest decorator

Gabrola:
Pretty neat little method to do "overloads" for arrow functions. This prevents the method decorator signature from being used on classes and vice-versa.

Screenshot 2023-02-02 172708


oliverbutler (approved):
This is great! I could see us getting a lot of issue reports without this

oliverbutler (commented):
Can you add a changeset please

PR #159, opened 2023-02-02
fix(core/client): wrong content-type detection.

### LINKED ISSUE #158: core/client: wrong "content-type" header detection
https://github.com/ts-rest/ts-rest/blob/main/libs/ts-rest/core/src/lib/client.ts#L113

There is not quite correct content-type checking.
Seems is better to check through "includes" otherwise it leads to such errors

image.

This results in an error where the request body is a blob.

evist0:
Closes #158

### DIFF
--- a/.changeset/hip-peaches-sell.md
+++ b/.changeset/hip-peaches-sell.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/core': patch
+---
+
+fix(core): wrong "content-type" detection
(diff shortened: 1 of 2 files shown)


oliverbutler (commented):
Nice catch! Could you please add a changeset with the command `pnpm changeset` as a patch, then lets get this merged. Thanks!

Gabrola (approved):
Looks good! Thanks for the fix!

PR #163, opened 2023-02-03
Document StrictNullChecks + other Doc enhancements

michaelangrivera:
fixes #162 

I recommend you spin up docusaurus locally to see the docs changes. I removed the redundant readme in the docs and had it point to the correct modules folder for the API reference. I also changed the sidebar layout order. 

Let me know what you think!

PR #165, opened 2023-02-05
docs: add popular package managers installation

rifaldhiaw:
Hi All, I really like this project and would love to contribute more to this project. I rarely contribute to an open-source project before. Please let me know if there is any step that I should do first before doing any PR. Such as, should I create an Issue first for this kind of Docs improvement? Thanks.   

Changes:
- Added tabs for popular package manager in the package installation section.
- Also, I fixed a little typescript type warning in `ts-rest-express.ts`.

https://user-images.githubusercontent.com/7936061/216814679-321bbfc2-db85-455d-867a-1e4781ae2963.mov

### DIFF
--- a/libs/ts-rest/express/src/lib/ts-rest-express.ts
+++ b/libs/ts-rest/express/src/lib/ts-rest-express.ts
@@ -15,9 +15,8 @@ import {
 } from '@ts-rest/core';
 import type {
   IRouter,
-  NextFunction,
   Request,
-  Response,
+  RequestHandler,
 } from 'express-serve-static-core';
 import type { IncomingHttpHeaders } from 'http';
 
@@ -171,7 +170,7 @@ const transformAppRouteQueryImplementation = (
 
       return res.status(statusCode).json(result.body);
     } catch (e) {
-      return next(e);
+      return next?.(e);
     }
   });
 };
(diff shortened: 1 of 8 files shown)


rifaldhiaw:
My bad, just realized that it might be in conflict with PR #163

oliverbutler:
This looks great! Thanks for the PR @rifaldhiaw

@all-contributors add @rifaldhiaw for documentation

oliverbutler (approved):

libs/ts-rest/express/src/lib/ts-rest-express.ts
```diff
@@ -171,7 +170,7 @@ const transformAppRouteQueryImplementation = (
 
       return res.status(statusCode).json(result.body);
     } catch (e) {
-      return next(e);
+      return next?.(e);
```
Gabrola, line 173: There used to be a problem with the typing of the `next` function because it was grabbing incorrect typing from `express-static-serve-core` instead of `@types/express-static-serve-core`, but that should be the case now.

As you can see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-serve-static-core/index.d.ts#L72) `next` is not optional, so there should be no need for the optional chaining operator

Gabrola:
@rifaldhiaw love this change. Great contribution mate! ♥

libs/ts-rest/express/src/lib/ts-rest-express.ts
rifaldhiaw: Ahh I see.., My bad, so It was because my `node_modules` wasn't up to date. I haven't done `pnpm install` again after syncing this [commit](https://github.com/ts-rest/ts-rest/commit/41a5fbbe45dc9563ebbf554ad0677b07997dab59#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L71).

Will revert it back in my next PR, thanks for the review!

PR #167, opened 2023-02-05
fix: do not throw error for missing content-type header

Gabrola:
According to [RFC 2616](https://www.rfc-editor.org/rfc/rfc2616#section-7.2.1) a response _SHOULD_, but not _MUST_, include a content-type header. There are cases where a content-type header is not required such as when returning an empty response. This is common for some DELETE endpoints that return a `204 No Content` response.

I think in this case, returning a blob is appropriate and let the developers handle it how they please

### DIFF
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -108,15 +108,11 @@ export const defaultApi: ApiFetcher = async ({
   const result = await fetch(path, { method, headers, body, credentials });
   const contentType = result.headers.get('content-type');
 
-  if (!contentType) {
-    throw TypeError('Resource Response missing content-type header');
-  }
-
-  if (contentType.includes('application/json')) {
+  if (contentType?.includes('application/json')) {
     return { status: result.status, body: await result.json() };
   }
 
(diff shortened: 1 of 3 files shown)


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -108,16 +108,14 @@ export const defaultApi: ApiFetcher = async ({
...
-    return { status: result.status, body: await result.json() };
-  }
+  if (contentType) {
+    if (contentType.includes('application/json')) {
+      return { status: result.status, body: await result.json() };
+    }
 
-  if (contentType.includes('text/plain')) {
-    return { status: result.status, body: await result.text() };
+    if (contentType.includes('text/plain')) {
```
oliverbutler, line 116: Is "contentType?." Nicer than the double nested if? 🤔

oliverbutler (approved):

libs/ts-rest/core/src/lib/client.ts
Gabrola: Absolutely! Still stuck in ES5 lol 🤦‍♂️

PR #169, opened 2023-02-05
feat: better OpenAPI generation and include zod pathParams

Gabrola:
Path params were included from the path itself only and not taking into account any extra zod typings provided through `pathParams`, so I've added that.

Nullish fields were also being generated as an incorrect

```javascript
{
  anyOf: [
    {
      not: {},
    },
    {
      type: 'string',
    },
  ],
}
```

The zod-openapi package fixes that

### DIFF
--- a/libs/ts-rest/open-api/src/lib/ts-rest-open-api.ts
+++ b/libs/ts-rest/open-api/src/lib/ts-rest-open-api.ts
@@ -5,7 +5,8 @@ import {
   OperationObject,
   PathsObject,
 } from 'openapi3-ts';
-import zodToJsonSchema from 'zod-to-json-schema';
+import { generateSchema } from '@anatine/zod-openapi';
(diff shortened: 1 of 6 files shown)


libs/ts-rest/open-api/src/lib/ts-rest-open-api.ts
```diff
@@ -40,50 +41,82 @@ const getPathsFromRouter = (
...
+      },
+    })) || [];
+
+  if (isZodObj) {
+    const paramsFromZod = Object.entries(zodObject.shape).map(
+      ([key, value]) => ({
+        name: key,
+        in: 'path' as const,
+        required: true,
+        schema: getOpenApiSchemaFromZod(value),
```
michaelangrivera, line 80: nice!

michaelangrivera (approved):
LGTM!

PR #177, opened 2023-02-19
enhance Custom Client example

### LINKED ISSUE #176: enhance Custom Client example
Our current example code does not handle errors correctly:

```
export class SampleAPI {
  token: string
  constructor(params: { token: string }) {
    this.token = params.token
    this.baseUrl = 'http://localhost:3333/api'
  }
  client = () => {
    return initClient(userContract, {
      baseUrl: this.baseUrl,
      baseHeaders: {
        Authorization: `Bearer ${idToken}`,
        'Content-Type': 'application/json'
      },
      api: async ({ path, method, headers, body }) => {
        const result = await axios.request({
          method: method as Method,
          url: `${this.baseUrl}/${path}`,
          headers,
          data: body,
        }) // <--- we should catch the axios error and pass it upstream for the user to handle
        return { status: result.status, body: result.data }
      },
    })
  }
}
```
React Query somehow recognizes an error happens (probably because the promise does not resolve). Still, when the error's data isn't passed, we cannot access any data (or typesafe data). 

We should signal users to catch and return data appropriately.

michaelangrivera:
fixes #176

### DIFF
--- a/apps/docs/docs/core/custom.md
+++ b/apps/docs/docs/core/custom.md
@@ -133,6 +133,3 @@ export class SampleAPI {
 ```
 
 :::
-
-
-   Authorization: `Bearer ${token}` 
--- a/apps/docs/docs/core/custom.md
+++ b/apps/docs/docs/core/custom.md
@@ -51,6 +51,7 @@ Sometimes you need dynamic headers, IE passing in a Bearer token. There are two 
 ```typescript
 import { contract } from './some-contract'
 import axios, { Method, AxiosError, AxiosResponse, isAxiosError } from 'axios'
+
 export class SampleAPI {
   token: string
   constructor(params: { token: string }) {
--- a/apps/docs/docs/core/custom.md
+++ b/apps/docs/docs/core/custom.md
@@ -110,7 +110,7 @@ export class SampleAPI {
         try {
           const result = await axios.request({
             method: method as Method,
-            url: `${this.serverEnvUrl}/${path}`,
+            url: `${this.baseUrl}/${path}`,
             headers: { 
               ...headers,
               Authorization: `Bearer ${idToken}` 
--- a/apps/docs/docs/core/custom.md
+++ b/apps/docs/docs/core/custom.md
@@ -38,8 +38,9 @@ const client = initClient(contract, {
         const response = error.response as AxiosResponse
         return { status: response.status, body: response.data }
       }
+      
+      throw e;
     }
-    return { status: 500, body: {} }
   },
 })
 ```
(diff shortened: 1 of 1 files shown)


apps/docs/docs/core/custom.md
```diff
@@ -15,20 +15,31 @@ you can pass a `api` attribute to the `initClient` or `initQueryClient`.
 
 **Here's a basic example: **
 ```typescript
+import { contract } from './some-contract'
+import axios, { Method, AxiosError, AxiosResponse, isAxiosError } from 'axios'
 const client = initClient(contract, {
```
Gabrola, line 20:
```suggestion
import axios, { Method, AxiosError, AxiosResponse, isAxiosError } from 'axios'

const client = initClient(contract, {
```

apps/docs/docs/core/custom.md
```diff
@@ -15,20 +15,31 @@ you can pass a `api` attribute to the `initClient` or `initQueryClient`.
...
+      })
+      return { status: result.status, body: result.data }
+    } catch (e: Error | AxiosError | any) {
+      if (isAxiosError(e)) {
+        const error = e as AxiosError
+        const response = error.response as AxiosResponse
+        return { status: response.status, body: response.data }
+      }
+    }
+    return { status: 500, body: {} }
```
Gabrola, line 42: I think it makes sense to re-throw the error if it is not an axios error

```suggestion
      }
      
      throw e;
    }
```

apps/docs/docs/core/custom.md
```diff
@@ -38,27 +49,38 @@ Sometimes you need dynamic headers, IE passing in a Bearer token. There are two
 ### Instantiate the `client` with the header passed in:
 
 ```typescript
+import { contract } from './some-contract'
+import axios, { Method, AxiosError, AxiosResponse, isAxiosError } from 'axios'
 export class SampleAPI {
```
Gabrola, line 54:
```suggestion
import axios, { Method, AxiosError, AxiosResponse, isAxiosError } from 'axios'

export class SampleAPI {
```

apps/docs/docs/core/custom.md
```diff
@@ -70,33 +92,47 @@ export class SampleAPI {
...
       },
     })
   }
 }
 ```
 
 :::
+
+
+   Authorization: `Bearer ${token}`
```
Gabrola, line 138: Accidental copy-pasta?

Gabrola (commented):
LGTM! Left a couple of nits

apps/docs/docs/core/custom.md
michaelangrivera: lol whoops

apps/docs/docs/core/custom.md
michaelangrivera: Yup, you're right; I was worried about the return type conflicting, but TS picks it up correctly.

oliverbutler (approved):

PR #179, opened 2023-02-21
Allow extra API client args, allow headers, and prettify query args

oliverbutler:
An action-packed PR! 

- [ ] Docs, I want to document the client args :) 

- Allow API client args to be set in the custom API fetcher
- Allow headers to be overridden
- Prettify query args (thanks @mattpocock) 

before
after

### DIFF
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -6,16 +6,17 @@ import {
   AreAllPropertiesOptional,
   Merge,
   OptionalIfAllOptional,
+  Prettify,
   Without,
   ZodInferOrType,
   ZodInputOrType,
 } from './type-utils';
 
-type RecursiveProxyObj<T extends AppRouter> = {
+type RecursiveProxyObj<T extends AppRouter, TClientArgs extends ClientArgs> = {
   [TKey in keyof T]: T[TKey] extends AppRoute
-    ? AppRouteFunction<T[TKey]>
+    ? AppRouteFunction<T[TKey], TClientArgs>
     : T[TKey] extends AppRouter
-    ? RecursiveProxyObj<T[TKey]>
+    ? RecursiveProxyObj<T[TKey], TClientArgs>
     : never;
 };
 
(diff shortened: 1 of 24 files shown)


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -133,23 +169,52 @@ const createFormData = (body: unknown) => {
...
+    combinedHeaders.set(key, value);
+  });
+
+  Object.entries(headers).forEach(([key, value]) => {
+    if (value === undefined) {
+      combinedHeaders.delete(key);
+    } else {
+      combinedHeaders.set(key, value);
+    }
+  });
```
Gabrola, line 208: Can we simplify this to something like this?

```suggestion
  const combinedHeaders = {
    ...clientArgs.baseHeaders,
    ...headers
  };
```

I'm not 100% sure how `fetch` would handle undefined header values, but I'd take a very strong guess that it simply omits the header.

Also we'd probably need to normalize the header names to lower case in either implementations, otherwise we can possibly set the same header twice without `headers` overwriting a header if the casing was different. I'd also guess if this happens, the second differently-cased header would overwrite the first internally, but I'd rather not depend on undefined behavior of `fetch`

michaelangrivera: Does header casing matter on the server side, though? I can't remember if frameworks like express and Nest auto-lowercase incoming headers?

Gabrola: @michaelangrivera it doesn't but this is mainly about JS object keys being case sensitive, in this case we can have weird behavior if the same header is passed twice but with different casing. To combat this, a lot of JS client-side libraries, including fetch, normalize all headers to lower case.

@oliverbutler I just tested it, and it seems that a repeated header with different name casing will append the values together.

```typescript
// baseHeaders = { 'content-type': 'application/json' }
// headers     = { 'Content-Type': 'text/plain' }
const combinedHeaders = {
  ...clientArgs.baseHeaders,
  ...headers
};
```

This will result in the header being set as `content-type: application/json, text/plain`

Gabrola (approved):
LGTM! Huge improvement to the docs!

PR #182, opened 2023-02-28
feat: add type-safe header definitions to contracts

### LINKED ISSUE #108: Add type-safe headers
I've had a couple of problems where I forgot to pass headers to my APIs, we could add this to add another level of type safety to ts-rest.

Gabrola:
Closes #108 (finally!)

RouterOptions should be expanded in a later PR to include `jsonQuery`, but leaving it out now to keep this PR as small as possible.

### Key Points:

- `baseHeaders` are merged into individual endpoint headers when `c.router(...)` is called. Doing this at this stage saves the hassle of creating complicated types and logic in each server/client library code; core/client, react-query, express, etc. are relatively untouched. Zod objects are merged, so we are able to use `z.infer` on the server and `z.input` on the client
- As a consequence of the above we cannot mix and match Zod schemas with plain `c.body`/`c.response` types; otherwise we'd need to maintain two different types `headersInput`/`headersOutput` where we merge the plain type once with the `z.infer` of the Zod object and once with the `z.input`. We'd also have this weird behavior where some headers will go through validation and some won't. I think this better, and this is clearly communicated by returning a detailed type error if Zod objects and types are mixed.
- Clients can pass extra headers that are not defined in the contract through `extraHeaders`. This can be useful for headers that might not be consumed by the application backend at all but by a reverse-proxy or something similar.
- On the server end all request headers including the ones not defined in the contract are combined into one object `headers`
- Nested contracts will not inherit options if used directly. This is intuitive though as the contracts are technically immutable, so it makes no sense that a contract includes anything from another (parent) contract that is not referenced at all in the code. See below:

```typescript
const postsContract = c.router({
  getPosts: {
    method: 'GET',
    path: '/posts',
    // ...
  },
});

const contract = c.router(
  {
    posts: postsContract,
  },
  {
    baseHeaders: z.object({
      'x-api-key': z.string(),
    }),
  }
);

const postsClient = initClient(postsContract, {
  baseUrl: 'https://api.com',
});

postsClient.getPosts({ ... });
// ^ 'x-api-key' header not added to headers type
```


libs/ts-rest/core/src/lib/client.spec.ts
```diff
@@ -541,25 +554,7 @@ describe('custom api', () => {
     await customClient.posts.getPost({
       params: { id: '1' },
```
Gabrola, line 555: We have a weird pre-existing bug here. For some reason when a custom client is used, all args become optional and args itself becomes optional. In this case, `params` should not be optional but the type for `getPost` is

`getPost: (args?: ({headers?: {"x-test"?: string | undefined, "base-header"?: string | undefined, "x-api-key"?: string | undefined} | undefined, params?: {id: string} | undefined, uploadProgress?: ((progress: number) => void) | undefined} | undefined)) => Promise<...>`

oliverbutler: I just played with this, I can replicate it, it looks like the "AreAllPropertiesOptional" work doesn't work correctly :(

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -70,14 +88,36 @@ type DataReturnArgsBase<
...
    * Unset a header by setting it to undefined
+   *
+   * Mandatory headers defined in the client's baseHeaders are optional
    */
-  headers?: Record<string, string>;
+  headers: Prettify<
+    Merge<
+      'baseHeaders' extends keyof TRouterOptions
+        ? PartialByLooseKeys<
+            ZodInputOrType<TRouterOptions['baseHeaders']>,
```
arivera-xealth, line 98: Nit: Can we change `baseHeaders` to maybe `defaultHeaders`? To me, `defaultHeaders` reads better

Gabrola: Yeah, I agree that the reads better! We already use `baseHeaders` for the client options, so not sure if it'd be optimal if we are inconsistent with our naming here

libs/ts-rest/core/src/lib/dsl.ts
```diff
@@ -29,25 +36,80 @@ export type AppRouteMutation = {
...
+  TOptions extends RouterOptions
+> = {
+  [TRouterKey in keyof TRouter]: TRouter[TRouterKey] extends AppRoute
+    ? Prettify<ApplyOptions<TRouter[TRouterKey], TOptions>>
+    : TRouter[TRouterKey] extends AppRouter
+    ? RecursivelyApplyOptions<TRouter[TRouterKey], TOptions>
+    : TRouter[TRouterKey];
+};
+
+type UniversalMerge<A, B> = A extends z.AnyZodObject
```
michaelangrivera, line 87: This literally blew my mind. Awesome. 

Need to spend a few days reviewing the whole PR to make sure I'm understanding everything fully! 😅

Gabrola: Yeah that gave me a bit of a headache 😅 but I think it was worth not having the headache of trying to figure out how to combine Zod objects with plain TS objects.

libs/ts-rest/core/src/lib/dsl.ts
```diff
@@ -61,14 +123,18 @@ export type AppRouter = {
...
+
 /**
  * Differentiate between a route and a router
  *
  * @param obj
  * @returns
  */
 export const isAppRoute = (obj: AppRoute | AppRouter): obj is AppRoute => {
-  return obj?.method !== undefined;
+  return 'method' in obj && 'path' in obj;
```
michaelangrivera, line 137: This makes sense, we'd always want method and path properties to exist.

libs/ts-rest/core/src/lib/client.spec.ts
```diff
@@ -100,29 +105,40 @@ const postsRouter = c.router({
...
+      responses: {
+        200: c.response<{ message: string }>(),
+      },
+      contentType: 'multipart/form-data',
     },
-    contentType: 'multipart/form-data',
   },
-});
+  {
+    baseHeaders: z.object({
```
michaelangrivera, line 129: So in this case, these baseHeaders will only apply to `upload` and `health`, but not `posts` correct?

libs/ts-rest/core/src/lib/client.spec.ts
```diff
@@ -528,16 +560,26 @@ const customClient = initClient(router, {
...
+        skip?: number;
+        order?: string;
+      };
+      headers: {
+        'x-pagination'?: string;
+        'x-test'?: string;
+        'base-header'?: string;
+        'x-api-key': string;
+      };
+      extraHeaders?: {
```
michaelangrivera, line 575: what are `extraHeaders`?

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -122,14 +132,12 @@ export function getRouteResponses<T extends AppRouter>(router: T) {
...
-> = AreAllPropertiesOptional<
-  Without<DataReturnArgs<TRoute, TClientArgs>, never>
-> extends true
+> = AreAllPropertiesOptional<DataReturnArgs<TRoute, TClientArgs>> extends true
   ? (
-      args?: Prettify<Without<DataReturnArgs<TRoute, TClientArgs>, never>>
+      args?: Prettify<DataReturnArgs<TRoute, TClientArgs>>
     ) => Promise<Prettify<ApiRouteResponse<TRoute['responses']>>>
   : (
-      args: Prettify<Without<DataReturnArgs<TRoute, TClientArgs>, never>>
```
michaelangrivera, line 132: Does this fix this comment? https://github.com/ts-rest/ts-rest/pull/182/files#r1119446639

libs/ts-rest/core/src/lib/type-utils.ts
```diff
@@ -103,3 +106,28 @@ export type Prettify<T> = {
...
+export type Opaque<Type, Token = unknown> = Type & Tagged<Token>;
+
+export type WithoutUnknown<T> = Pick<
+  T,
+  {
+    [K in keyof T]: unknown extends Exclude<T[K], undefined> ? never : K;
+  }[keyof T]
+>;
+
+export type LowercaseKeys<T> = Prettify<{
```
michaelangrivera, line 131: Since we're doing Lowercase type casting for header keys, where are we ensuring that the header keys are lowercase? Is it after you initialize the `router` or during the consumption of the header (in say, an express server endpoint for instance)?

libs/ts-rest/express/src/lib/ts-rest-express.ts
```diff
@@ -117,6 +118,45 @@ const recursivelyApplyExpressRouter = (
...
+  const paramsResult = checkZodSchema(req.params, schema.pathParams, {
+    passThroughExtraKeys: true,
+  });
+
+  if (!paramsResult.success) {
+    return res.status(400).send(paramsResult.error);
+  }
+
+  const headersResult = checkZodSchema(req.headers, schema.headers, {
+    passThroughExtraKeys: true,
```
michaelangrivera, line 136: This should be an option IMO. Can we let the developer choose whether or not they want to `passThroughExtraKeys`?

libs/ts-rest/express/src/lib/ts-rest-express.ts
```diff
@@ -117,6 +118,45 @@ const recursivelyApplyExpressRouter = (
...
+  }
+
+  const query = options.jsonQuery
+    ? parseJsonQueryObject(req.query as Record<string, string>)
+    : req.query;
+
+  const queryResult = checkZodSchema(query, schema.query);
+
+  if (!queryResult.success) {
+    return res.status(400).send(queryResult.error);
```
michaelangrivera, line 150: We'll want to document what these responses will look like when writing the documentation. Most devs will want to know that we'll send a 400 error (and the error string) if the validation does not pass.

libs/ts-rest/express/src/lib/ts-rest-express.ts
michaelangrivera: Also, don't we want to parse the headers first, before checking the params?

libs/ts-rest/core/src/lib/client.spec.ts
michaelangrivera: Is this to allow the developer to pass extra headers during runtime? 
How is this handled on the express/nestjs side?

Gabrola: Yes, exactly! I forgot to mention that in the comments. On the server they will just be merged into the `req.headers`/`params.headers`. It might make sense to separate them on the server too, I'm not sure.

libs/ts-rest/core/src/lib/client.ts
Gabrola: Nah, that was fixed in a separate PR. This is just cleaning up the type to reduce duplication. The `Without` part was moved into `DataReturnArgs`

libs/ts-rest/core/src/lib/type-utils.ts
Gabrola: Express `req.headers` is always normalized in lowercase so that's why we need to ensure the accessor types are in lowercase too on the server-side. On the client side it wouldn't really matter if we weren't merging types, but since we do, if we had the same header with different casing in `baseHeaders` and an endpoint's `headers` they would be included together as two separate properties.

libs/ts-rest/express/src/lib/ts-rest-express.ts
Gabrola: In terms of the order, I'm thinking of it in a top-down fashion when observing a request. In my head it goes: URL (path params) -> headers -> query -> body

libs/ts-rest/express/src/lib/ts-rest-express.ts
Gabrola: Yeah, I agree, that makes sense. This is the current behavior already so we'd have to address that in a separate PR.

However, I think the right thing to do would be to do a small refactor to the express implementation where we throw an error, and use a ts-rest-supplied error handler middleware to send the standard zod error that can also be swapped out so the application can send custom error messages.

Gabrola: We can technically already do this on the NestJS implementation by using an exception filter and catching `BadRequestException`

libs/ts-rest/express/src/lib/ts-rest-express.ts
```diff
@@ -45,7 +46,7 @@ type AppRouteQueryImplementation<T extends AppRouteQuery> = (
     {
       params: PathParamsWithCustomValidators<T>;
       query: ZodInferOrType<T['query']>;
-      headers: IncomingHttpHeaders;
+      headers: LowercaseKeys<ZodInferOrType<T['headers']>> & Request['headers'];
```
oliverbutler, line 49: Nice - much improved with lowercase keys :)

oliverbutler (approved):
This looks awesome. Still not 100% on the extraHeaders as tradeoff for improved intellisense, regardless (chat in Discord), however, I love this.

RWOverdijk:
What about response headers?

Gabrola:
@RWOverdijk yes, we'll work on that soon, but it won't be part of the contract for the time being

RWOverdijk:
@Gabrola Ah, cool! That's fair, I was just checking if I didn't misunderstand 😄

PR #183, opened 2023-03-01
fix: broken typing for optional arguments

Gabrola:
Fixes this issue where args was optional despite some args being required

Screenshot 2023-03-01 082111

### DIFF
--- a/libs/ts-rest/core/src/lib/type-utils.ts
+++ b/libs/ts-rest/core/src/lib/type-utils.ts
@@ -40,11 +40,11 @@ export type GetFieldType<T, P> = P extends `${infer Left}.${infer Right}`
 // https://stackoverflow.com/questions/63447660/typescript-remove-all-properties-with-particular-type
 // Nested solution also available ^
 type ExcludeKeysWithTypeOf<T, V> = {
-  [K in keyof T]: Exclude<T[K], undefined> extends V ? never : K;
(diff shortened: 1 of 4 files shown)


libs/ts-rest/core/src/lib/type-utils.ts
```diff
@@ -40,11 +40,11 @@ export type GetFieldType<T, P> = P extends `${infer Left}.${infer Right}`
 // https://stackoverflow.com/questions/63447660/typescript-remove-all-properties-with-particular-type
 // Nested solution also available ^
 type ExcludeKeysWithTypeOf<T, V> = {
-  [K in keyof T]: Exclude<T[K], undefined> extends V ? never : K;
+  [K in keyof T]-?: Exclude<T[K], undefined> extends V ? never : K;
```
Gabrola, line 43: @oliverbutler `Without<DataReturnArgs, never>` was not working and returning `unknown` because the new `headers` argument is optional and these mapped types did not include optional keys. This is what was triggering everything to become optional.

oliverbutler (approved):
Amazing, love the use of the typescript testing, we should use that way more haha

Gabrola:
> Amazing, love the use of the typescript testing, we should use that way more haha

@oliverbutler yeah, thanks for adding that! Doing it this way is much less clunky than that expect-type package imo

PR #189, opened 2023-03-24
fix: restore getRouteResponses method

DylanSimowitz:
The getRouteResponses method was removed (mistakenly?), but provides a nice utility function for the client to access response types. The method is still documented here https://ts-rest.com/docs/core/#response-types

### DIFF
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -273,6 +273,21 @@ export const getRouteQuery = <TAppRoute extends AppRoute>(
   };
 };
 
+export type ApiResponseForRoute<T extends AppRoute> = ApiRouteResponse<
+  T['responses']
+>;
+
+// takes a router and returns response types for each AppRoute
+// does not support nested routers, yet
+
+export function getRouteResponses<T extends AppRouter>(router: T) {
+  return {} as {
+    [K in keyof typeof router]: typeof router[K] extends AppRoute
+      ? ApiResponseForRoute<typeof router[K]>
+      : 'not a route';
+  };
+}
+
 export type InitClientReturn<
   T extends AppRouter,
   TClientArgs extends ClientArgs


Gabrola:
Hey @DylanSimowitz thank you for your contribution! If I remember correctly, this function was used exclusively by the Nest library, hence why it was removed from the core client. In order to provide this functionality publicly we'll need to distinguish between inferring the Input schema and Output schema for Zod depending on whether the type will be used by the client or server; kind of similar to how trpc has `inferRouterInputs` and `inferRouterOutputs`.

For example for responses, we'd need to use the Zod Output on the client, and Zod Input on the server. For everything else (query, body, etc.) it's vice-versa.

Also I'd say using a TS Type (to be used like `GetRouteResponses<typeof contract.endpoint>`) rather than a function would be more appropriate for this functionality.

Gabrola:
Hey @DylanSimowitz thank you for your contribution! If I remember correctly, this function was used exclusively by the Nest library, hence why it was removed from the core client. In order to provide this functionality publicly we'll need to distinguish between inferring the Input schema and Output schema for Zod depending on whether the type will be used by the client or server; kind of similar to how trpc has `inferRouterInputs` and `inferRouterOutputs`.

For example for responses, we'd need to use the Zod Output on the client, and Zod Input on the server. For everything else (query, body, etc.) it's vice-versa.

Also I'd say using a TS Type (to be used like `GetRouteResponses<typeof contract>`) rather than a function would be more appropriate for this functionality.

michaelangrivera:
> Hey @DylanSimowitz thank you for your contribution! If I remember correctly, this function was used exclusively by the Nest library, hence why it was removed from the core client. In order to provide this functionality publicly we'll need to distinguish between inferring the Input schema and Output schema for Zod depending on whether the type will be used by the client or server; kind of similar to how trpc has `inferRouterInputs` and `inferRouterOutputs`.
> 
> For example for responses, we'd need to use the Zod Output on the client, and Zod Input on the server. For everything else (query, body, etc.) it's vice-versa.
> 
> Also I'd say using a TS Type (to be used like `GetRouteResponses<typeof contract>`) rather than a function would be more appropriate for this functionality.

Although it was only used by the Nest library directly, it was actually made to be general so that any client can import and use it. It was added to the core so that developers have a utility function for getting types from their router. 

I can work on adding this back, although as @Gabrola  said, this should be a type that the caller can import.

michaelangrivera:
@DylanSimowitz I will take this and fix it for you!

michaelangrivera (approved):

PR #192, opened 2023-03-31
restore getRouteResponses (for now), add ResponsesForRouter

### LINKED ISSUE #191: getRouteResponses accidentally deprecated
as mentioned in #189 , `getRouteResponses` was accidentally removed without warning. 

We need to restore this method and deprecate it officially while also providing new/similar functionality to existing users.

michaelangeloio:
fixes #191 

@Gabrola mentions that 

>  In order to provide this functionality publicly we'll need to distinguish between inferring the Input schema and Output schema for Zod depending on whether the type will be used by the client or server; kind of similar to how trpc has inferRouterInputs and inferRouterOutputs.

> For example for responses, we'd need to use the Zod Output on the client, and Zod Input on the server. For everything else (query, body, etc.) it's vice-versa.

I do eventually think we should take this idea and run with it eventually, however, for the sake of brevity, we need to do the following: 
- we should restore the original method
- add a "better" way of using the original method
- document the release versions that this is affected by

We should be careful moving forward on removing methods without proper communication/documentation.

### DIFF
--- a/apps/docs/docs/core/core.md
+++ b/apps/docs/docs/core/core.md
@@ -119,19 +119,14 @@ export const contract = c.router({
 
 ## Response Types
 
-If you need to quickly extract the Response types from a given Contract, you can use `getRouteResponses`.
-
-:warning: This does not support nested contracts yet. You need to pass in a defined contract. 
-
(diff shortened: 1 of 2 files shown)


oliverbutler (approved):
Nice! Agreed on the point about being careful with exposed methods 💯

PR #193, opened 2023-04-02
feat: add type inference helpers

Gabrola:
Addressing the issues raised in #192 with the extra added benefit of being able to handle nested contracts.

Given some extra scrutiny, this code can be used to deduplicate a lot of code in the client libraries and also the server libraries.

### DIFF
--- a/apps/docs/docs/core/infer-types.md
+++ b/apps/docs/docs/core/infer-types.md
@@ -1,0 +1,69 @@
+# Inferring Types
+
+Often, we need to manually extract the request or responses types of specific contract endpoints, so functions, services, lambdas, React components, etc. can be safely typed
+when there is no automatic type inference.
+
+We have separate type helpers for server-side and client-side code since we need to infer either the Input or Output Zod types depending on
+where the code is used.
+
+## Inferring Response Types
+
+To get the response types of a contract or a specific endpoint, we have the following type helpers:
+
+- `InferResponsesForServer<AppRouter | AppRoute, OptionalHttpStatusCode>`
+- `InferResponsesForClient<AppRouter | AppRoute, OptionalHttpStatusCode>`
+
+```typescript
+import { InferResponsesForServer } from '@ts-rest/core';
+import { contract } from './contract';
+
+type ResponseShapes = InferResponsesForServer<typeof contract>;
+
+async function someHttpCall(req: Request): Promise<ResponseShapes['getPosts']> {
+  return ...;
+}
+
+function someServiceCall(): InferResponsesForServer<typeof contract.getPosts> {
+  return ...;
+}
+```
+
+### Inferring Response Body
+
+If you need to infer the response body for a defined response status of a specific endpoint, we can use the following type helpers:
+
+- `InferResponseBodyForServer<AppRoute, OptionalHttpStatusCode>`
+- `InferResponseBodyForClient<AppRoute, OptionalHttpStatusCode>`
+
+This is syntactic sugar for `InferResponsesForServer<AppRoute, OptionalHttpStatusCode & keyof AppRoute['responses']>['body']`
+
+```typescript
+import React from 'react';
+import { InferResponseBodyForClient } from '@ts-rest/core';
+import { contract } from './contract';
+
+type Post = InferResponseBodyForClient<typeof contract.getPost, 200>;
+
+function PostComponent(props: { post: Post }) {
+  return <>...</>;
+}
+```
+
+## Inferring Request Types
+
+To get the request (path params, query params, body) types of a contract or a specific endpoint, we have the following type helpers:
+
+- `InferRequestForServer<AppRouter | AppRoute>`
+- `InferRequestForClient<AppRouter | AppRoute>`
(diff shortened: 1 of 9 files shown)


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -91,24 +98,20 @@ export type ApiRouteResponse<T> =
       body: unknown;
     };
 
-export type ResponseForRoute<T extends AppRoute> = ApiRouteResponse<
+/**
+ * @deprecated Only safe to use on the client-side. Use `InferResponsesForServer`/`InferResponsesForClient` instead.
+ */
+export type ApiResponseForRoute<T extends AppRoute> = ApiRouteResponse<
```
Gabrola, line 104: @michaelangrivera `ApiResponseForRoute` was the old name of the type used by `getRouteResponses` previously so I've renamed it back to that. This should be fine since the last merged commits have still not being released.

apps/docs/docs/core/infer-types.md
```diff
@@ -0,0 +1,69 @@
...
+- `InferResponseBodyForClient<AppRoute, OptionalHttpStatusCode>`
+
+This is syntactic sugar for `InferResponsesForServer<AppRoute, OptionalHttpStatusCode & keyof AppRoute['responses']>['body']`
+
+```typescript
+import React from 'react';
+import { InferResponseBodyForClient } from '@ts-rest/core';
+import { contract } from './contract';
+
+type Post = InferResponseBodyForClient<typeof contract.getPost, 200>;
```
michaelangrivera, line 45: Instead of `InferResponseBodyForClient`, why not add another type generic param that allows the dev to pass in `server` or `client` and rename to `InferResponseBody`? 

So it would look like this:
```suggestion
type Post = InferResponseBody<typeof contract.getPost, 200, 'server'>;
```

libs/ts-rest/core/src/lib/client.ts
michaelangrivera: LGTM!

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -91,24 +98,20 @@ export type ApiRouteResponse<T> =
...
-  [K in keyof T]: T[K] extends AppRoute
-  ? ResponseForRoute<T[K]>
-  : T[K] extends AppRouter ? ResponsesForRouter<T[K]> : never;
-};
+>;
 
 /**
- * 
- * @deprecated
+ * @deprecated Only safe to use on the client-side. Use `InferResponsesForServer`/`InferResponsesForClient` instead.
```
michaelangrivera, line 109: 👏

libs/ts-rest/core/src/lib/infer-types.ts
```diff
@@ -0,0 +1,128 @@
...
+import { HTTPStatusCode } from './status-codes';
+import {
+  Prettify,
+  Without,
+  ZodInferOrType,
+  ZodInputOrType,
+} from './type-utils';
+import { PathParamsWithCustomValidators } from './client';
+
+type AppRouteResponses<
```
michaelangrivera, line 11: Do we not want to expose `AppRouteResponses`? In case the developer/consumer would like to play around with it?

libs/ts-rest/core/src/lib/infer-types.ts
```diff
@@ -0,0 +1,128 @@
...
+            query: 'query' extends keyof T ? ZodInferOrType<T['query']> : never;
+          },
+          never
+        >
+      >
+    : T extends AppRouter
+    ? {
+        [TKey in keyof T]: T[TKey] extends AppRoute
+          ? InferRequestForServer<T[TKey]>
+          : T[TKey] extends AppRouter
```
michaelangrivera, line 93: I really like the usage of `TKey`, it is a lot easier to read than the typical `T in keyof K` that you usually see in TS generics

apps/docs/docs/core/infer-types.md
Gabrola: @michaelangrivera For the code on our end, yes, it would help a bit with code reuse. I think from a developers perspective though that it's safer to have separate types with `Client` or `Server` in the type name; it'll be clearer in terms of readability and reduces the chances of copy-pasting errors from being missed. To hammer on that point, I actually changed it now to `ServerInfer` and `ClientInfer`, so now it's shorter (I was also annoyed with `ForClient`/`ForServer`) and also much more readable.

libs/ts-rest/core/src/lib/infer-types.ts
Gabrola: I think it's a double edged sword. For example exposing something like `ZodInferOrType` allowed developers to build what is essentially this entire PR without having to depend on us adding this functionality, but at the same time it makes backwards compatibility harder for us if we want to do any refactors.

I think a lot of our internal functions and types are being "involuntarily" exposed because we need them to be exported so they can be used internally by other typescript files or other libraries but not because we actually intended to expose them.

PR #199, opened 2023-04-10
Add rawBody and contentType to Custom Client API arguments

### LINKED ISSUE #198: Custom Client automatically stringifies body and does not provide access to raw body object
Take the following example from our docs:
```typescript
import { contract } from './some-contract';
import axios, { Method, AxiosError, AxiosResponse, isAxiosError } from 'axios';

export class SampleAPI {
  token: string;
  constructor(params: { token: string }) {
    this.token = params.token;
    this.baseUrl = 'http://localhost:3333/api';
  }
  client = () => {
    return initClient(contract, {
      baseUrl: this.baseUrl,
      baseHeaders: {
        Authorization: `Bearer ${idToken}`,
        'Content-Type': 'application/json',
      },
      api: async ({ path, method, headers, body }) => {
        try {
          const result = await axios.request({
            method: method as Method,
            url: `${this.baseUrl}/${path}`,
            headers,
            data: body,
          });
          return { status: result.status, body: result.data };
        } catch (e: Error | AxiosError | any) {
          if (isAxiosError(e)) {
            const error = e as AxiosError;
            const response = error.response as AxiosResponse;
            return { status: response.status, body: response.data };
          }
          throw e;
        }
      },
    });
  };
}
```
The `body` passed here is either the `FormData`, `string` , `undefined` or `null` type. We are automatically stringifying the body before passing it to the custom `api` client as shown here https://github.com/ts-rest/ts-rest/blob/81cfaddbcb9cdef0da23b54dd14c1c4ba029bf74/libs/ts-rest/core/src/lib/client.ts#L252. 

Auto-stringifying the body works fine for our `tsRestFetchApi` implementation but does not provide enough flexibility to developers who need access to the raw body. Ideally, when a developer chooses to create a custom client, they should be able to opt-in to auto-stringifying the body for POST/PUT requests. 

We _can_ pass in extra arguments of course, but we should have a dedicated key to guide developers _and_ also allow developers to disable that stringify operation (if needed).

michaelangrivera:
This aims to partially fix #198. 

My coworker discovered this issue when working with API Gateway/AWS Amplify. We have to do the following custom logic when using a Custom Client:

```typescript
// Convert body back to JSON before sending. Right now ts-rest will always stringify the body.
// This causes issue when deployed with API Gateway as API Gateway will also stringify the body.
result = await this.api.post(appConfig.APP_BASENAME, path, { body: JSON.parse(body as string), response: true })
```

The above means the ts-rest custom client stringifies the body, and then we have to _parse_ that stringified body for it to be _re-stringified_ again. This is a big no-no when it comes to performance. 

This PR adds `rawBody` and `contentType` to the params passed to the Custom Client. I know we can use the `extraInputArgs` feature. However, I think developers should have more guidance when developing Custom Clients, and adding these extra runtime params/types should help. 

Eventually, developers _should_ be able to opt out of that extra `JSON.stringify` operation if they need to for performance reasons. 

These are simply additions so that we do not introduce any breaking changes. In 4.0 we should revisit the implementation of this section of code.

### DIFF
--- a/apps/docs/docs/core/custom.md
+++ b/apps/docs/docs/core/custom.md
@@ -44,22 +44,6 @@ const client = initQueryClient(postsApi, {
 });
 ```
 
-### Accessing the raw body and content type
-
-Ts-Rest automatically stringifies your `body` input for mutations. You can access the raw body object and content-type like so:
-
-```typescript
-const client = initQueryClient(postsApi, {
-  baseUrl: 'http://localhost:5003',
-  baseHeaders: {},
-  api: async ({ path, method, headers, body, rawBody, contentType }) => {
-    // do something with rawBody ✨
-    return tsRestFetchApi(args);
-  },
-});
-```
-
-
 The magical bit, is this is now fully typed and will work with your IDE's autocomplete! 🤯
 
 One note here, any extra args which are provided here but aren't typed correctly - e.g. if you've `@ts-expect-error`'d, **will still be passed to your api**. This is because the `args` parameter is a spread of all the other arguments you pass in to your api.
@@ -75,6 +59,21 @@ client.getPosts({
 :::tip
 You can use this to accomplish loads of patterns, such as adding a `cache` argument to your api, or adding a `logger` argument to your api - maybe you want to add an `onUploadProgress` argument to your api to track upload progress? You can do all of this with the `args` parameter!
 :::
+
+### Accessing the raw body and content type
+
+Ts-Rest automatically stringifies your `body` input for mutations. You can access the raw body object and content-type like so:
+
+```typescript
+const client = initQueryClient(postsApi, {
+  baseUrl: 'http://localhost:5003',
+  baseHeaders: {},
+  api: async ({ path, method, headers, body, rawBody, contentType }) => {
+    // do something with rawBody ✨
+    return tsRestFetchApi(args);
+  },
+});
+```
 
 ## Using Axios (custom api override)
 
--- a/apps/docs/docs/core/custom.md
+++ b/apps/docs/docs/core/custom.md
@@ -46,7 +46,7 @@ const client = initQueryClient(postsApi, {
 
 ### Accessing the raw body and content type
 
-Our client automatically stringifies your `body` input for mutations. You can access the raw body object and content-type like so:
+Ts-Rest automatically stringifies your `body` input for mutations. You can access the raw body object and content-type like so:
 
 ```typescript
 const client = initQueryClient(postsApi, {
--- a/apps/docs/docs/core/custom.md
+++ b/apps/docs/docs/core/custom.md
@@ -44,6 +44,22 @@ const client = initQueryClient(postsApi, {
 });
 ```
 
+### Accessing the raw body and content type
+
+Our client automatically stringifies your `body` input for mutations. You can access the raw body object and content-type like so:
+
+```typescript
+const client = initQueryClient(postsApi, {
+  baseUrl: 'http://localhost:5003',
+  baseHeaders: {},
+  api: async ({ path, method, headers, body, rawBody, contentType }) => {
+    // do something with rawBody ✨
+    return tsRestFetchApi(args);
+  },
+});
+```
+
+
 The magical bit, is this is now fully typed and will work with your IDE's autocomplete! 🤯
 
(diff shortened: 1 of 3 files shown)


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -250,6 +254,8 @@ export const fetchApi = ({
     },
     body:
       body !== null && body !== undefined ? JSON.stringify(body) : undefined,
+    rawBody: body,
+    contentType: route.method !== 'GET'  ? route.contentType : undefined,
```
Gabrola, line 258: In mutations, `route.contentType` can be undefined, and when it is so, it is assumed to be `application/json`, but in this case we are still passing `undefined` here.

Since at the moment `contentType` can be `'application/json' | 'multipart/form-data' | undefined`, and the form-data case is already handled in the above code, we can hard code this to `application/json` just like on line 252 (we are also forcing the body to be `JSON.stringify`-ied so it can't be anything other than `application/json` at the moment)

Gabrola (changes requested):
Nice one! 🔥

libs/ts-rest/core/src/lib/client.ts
Gabrola: I think it might also make sense to change that line 252 thing to `...(route.method !== 'GET' && { 'content-type': 'application/json' })`, it doesn't really make sense to send a 'content-type' header with a GET request.

michaelangrivera: Ah, so maybe make the `contentType` an optional property then?

apps/docs/docs/core/custom.md
```diff
@@ -44,6 +44,22 @@ const client = initQueryClient(postsApi, {
...
+  baseUrl: 'http://localhost:5003',
+  baseHeaders: {},
+  api: async ({ path, method, headers, body, rawBody, contentType }) => {
+    // do something with rawBody ✨
+    return tsRestFetchApi(args);
+  },
+});
+```
+
+
```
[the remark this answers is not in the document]
michaelangrivera: Will fix!

libs/ts-rest/core/src/lib/client.ts
Gabrola: @michaelangrivera seems like it's already optional since it's inheriting from `AppRouteMutation['contentType']` which is also optional. But I think we should only pass `undefined` if method is GET, otherwise it must always be set. The core library is allowing undefined in the contract definition, but it assumes that as `application/json`, so we should also expose it as such to the custom client.

michaelangrivera: Makes sense! I will update!

michaelangrivera: Actually, `contentType` is only inheriting the _types_ from `AppRouteMutation['contentType']`, not the optional property key itself, so `contentType` is _not_ marked as optional, so we are OK in this sense (using the types from `AppRouteMutation['contentType']` for now). 

I've changed the value to always be `application/json` or `multipart/form-data`, though. Please take a look!

Gabrola: Looks good!

And yeah, my bad on using "optional" and having a type of a undefined interchangeably. I meant that the key in `AppRouteMutation` being optional makes the type of `AppRouteMutation['contentType']` include undefined in its union type. So yes, the new param `contentType` is not optional (rightly so), but we can still pass undefined to it in the case of GET requests.

michaelangrivera:
> Hello @michaelangrivera ,
> 
> Is breaking this really a breaking change? How many users are really using custom fetchers and relying on body to be a string (meaning custom fetchers not being implemented with axios or that already transform objects to string?) My humble guess is zero. But maybe i'm wrong. So here i would break it and remove the stringify and do it in a fetch adapter that check parameters
> 
> ```ts
> if (typeof config.body === "object") {
>     headers.set("Content-Type", "application/json");
>     headers.set("Accept", "application/json");
>     config.body = JSON.stringify(config.body);
>   }
> ```
> 
> See an example implementation here: https://github.com/ecyrbe/zodios/blob/feature/v11/packages/fetch/src/fetch-provider/fetch.ts

I think we plan to revisit this entire implementation in the future, so I will definitely note this down for then. I definitely agree, though.

michaelangrivera:
@Gabrola I think it's best we also merge the other non-breaking PRs too and release them in one update!

PR #200, opened 2023-04-12
fix: compatibility with Node.js TS native ES modules

### LINKED ISSUE #188: node16 (ESM) module resolution doesn't work
Hi there, 

Using @ts-rest in an ESM project that uses moduleResolution "node16" breaks typesafety:

CleanShot 2023-03-16 at 14 25 12@2x

### Additional information 

The project set up is overly complicated, making it harder to contribute.

- @ts-rest is "type: module" but the tsconfig in core uses commonjs for module resolution
- the exports field references a different declaration file `./src/index.d.ts` instead of the declaration file (which is not created, why not?) of `./index.js`
- npm does not work, pnpm is required
- `pnpm i` takes several minutes 
- the nx setup seems overkill for a simple library

Gabrola:
Fixes #188 

NX was adding `"type": "module"` to the compiled package.json files because we compile `esm` format JS files. This package.json field makes the Typescript compiler of consuming code think that ts-rest libraries are native ES modules, which in fact they are not. We compile both ESM and CJS JS files, but the compiled .d.ts files are in CJS format only, so it needs to be indicated correctly which format the project originally is, so Typescript knows how to transpile our definition files.

### DIFF
--- a/.changeset/moody-years-fail.md
+++ b/.changeset/moody-years-fail.md
@@ -1,0 +1,11 @@
+---
+'@ts-rest/core': patch
+'@ts-rest/express': patch
+'@ts-rest/nest': patch
+'@ts-rest/next': patch
+'@ts-rest/open-api': patch
+'@ts-rest/react-query': patch
+'@ts-rest/solid-query': patch
+---
+
+Fix comptability with Node.js TS native ESM code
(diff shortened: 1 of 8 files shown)


michaelangrivera (approved):
LGTM!!

PR #206, opened 2023-04-19
feat: export ts-rest react query types

Gabrola:


libs/ts-rest/react-query/src/index.ts
```diff
@@ -1 +1,2 @@
 export * from './lib/react-query';
+export * from './lib/types';
```
michaelangrivera, line 2: are we sure we want to expose _all_ of the types?

michaelangrivera (approved):

libs/ts-rest/react-query/src/index.ts
Gabrola: Inside the `types.ts` file, only the types that are needed are exported so we should be good here 👍

PR #209, opened 2023-05-07
Filter out undefined values from json queries

gingermusketeer:
If undefined values are present they are converted into the string `undefined`. This is not a valid JSON value so it fails parsing at [parseJsonQueryObject](https://github.com/ts-rest/ts-rest/blob/8011ec7c3cf098570d9f266a4f4cde4a36e38f82/libs/ts-rest/core/src/lib/query.ts#L104).

Filtering it out on the client side means that it will still have the value of undefined on the server.

P.S Thanks for a great library!!

### DIFF
--- a/.changeset/pretty-cows-shave.md
+++ b/.changeset/pretty-cows-shave.md
@@ -1,0 +1,5 @@
+---
+"@ts-rest/core": patch
+---
+
+Filter out undefined values from json queries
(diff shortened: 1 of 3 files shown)


Gabrola:
Ah, the infamous `JSON.parse(JSON.stringify(undefined))` error. Good catch, especially in our case, since `parseJsonQueryObject` would fall back to the string `undefined` rather than throwing an error. Thank you!

Gabrola (approved):
Thank you for the contribution! I made a small nit optimization to the code, otherwise looks great!

gingermusketeer:
Thank you!!

PR #211, opened 2023-05-09
feat: OpenAPI support Zod Refine

oliverbutler:
This was missing and threw and error if you tried to instantiate a schema with a refined object at the top level!


Gabrola (approved):
Looks good! 👍

PR #216, opened 2023-05-18
feat: react-query typed QueryClient function wrappers

### LINKED ISSUE #208: Allow access to raw query in custom api client
I'm running into a use-case where I need to have access to the raw query parameter object when using a custom client (axios in this example). I'm interfacing with an external API, so I'm unable to implement these fixes in the backend. 

---

My use-case is as follows:
1. The backend I'm interfacing with only supports unordered array query strings
```
# Does not work (indices array)
?item[0]=000&item[1]=111

# Does not work (brackets array)
?item[]=000&item[]=111

# Does not work (json)
?item=["000","111"]

# Works (repeated array)
?item=000&item=111
```

2. The api client only gives us access to `path` which returns the full path and serialized query string which returns the properties with array indices. ~~I could probably work around this by creating a custom function to re-parse the already serialized query string but that is not an ideal solution.~~
https://github.com/ts-rest/ts-rest/blob/d11071d20dd97a950755404f0d94fcfeab97672b/libs/ts-rest/core/src/lib/query.ts#L70-L72

Edit -- My current workaround is to parse the query string from the path using `qs`.
```typescript
const parsePath = (fullPath: string) => {
  const [path, params] = fullPath.split('?');

  const parsedParams = qs.parse(params);

  return {
    params: parsedParams,
    path,
  };
};
```

---

With access to the query object, I would then be able to pass the query object into the axios params directly and add my own serialization, e.g.:

```javascript
axios.defaults.paramsSerializer = (params) => {
  return qs.stringify(params, { arrayFormat: 'repeat' });
};
```

Could this be added?
Thank you!

Gabrola:
Closes #208 too

Bumped minimum @tanstack/react-query version to 4.18.0 (released 6 months ago) since `ensureQueryData` is not available in previous versions. Technically, you can get away with not installing a compatible version as long as you don't call our `ensureQueryData` helper


libs/ts-rest/react-query/src/lib/react-query.ts
```diff
@@ -106,23 +137,13 @@ type DataReturnQueries<
...
-> = AreAllPropertiesOptional<
-  Without<DataReturnArgs<TAppRoute, TClientArgs>, never>
-> extends true
-  ? (
-      queryKey: QueryKey,
-      args?: (
-        context: QueryFunctionContext<QueryKey>
-      ) => Without<DataReturnArgs<TAppRoute, TClientArgs>, never>,
-      options?: UseInfiniteQueryOptions<TAppRoute>
-    ) => UseInfiniteQueryResult<TAppRoute>
```
Gabrola, line 118: Removed the optional version as it makes no sense to use an infinite query with no args, as you'd need to at least map your page number to some arg.

libs/ts-rest/react-query/src/lib/react-query.ts
```diff
@@ -106,23 +137,13 @@ type DataReturnQueries<
 type DataReturnInfiniteQuery<
   TAppRoute extends AppRoute,
   TClientArgs extends ClientArgs
-> = AreAllPropertiesOptional<
-  Without<DataReturnArgs<TAppRoute, TClientArgs>, never>
-> extends true
-  ? (
-      queryKey: QueryKey,
-      args?: (
```
Gabrola, line 114: Removed the optional version as it makes no sense to use an infinite query with no args, as you'd need to at least map your page number to some arg. Technically a breaking change, but extremely unlikely that anyone is doing this.

michaelangeloio (approved):

oliverbutler (approved):
LGTM!

michaelangeloio (approved):

PR #228, opened 2023-05-28
Implement strict mode to throw on unexpected responses

### LINKED ISSUE #220: Restrict returned status codes to those that are defined
Been using this library for a month or so and have really enjoyed it. Has been a dramatic improvement for the flow of work!

One gotcha that I  have run into a few times is checking the response status. At the moment all the documented HTTP statuses are considered possible in TS. Would you be open to restricting these to only the defined statuses?

For example if you defined 200 as the only response status then you don't need to check the status field. If you defined two then checking for one would narrow both code paths.

This could be achieved via a flag to raise exceptions for non defined statuses.

## Current state

```ts
const postsContact = c.router({
  update: {
    method: "PUT",
    path: "/posts/:postId",
    body: PostSchema,
    responses: {
      200: ResponseSchema,
    },
    summary: "Update post information",
  },
});
const tsRestClient = initClient(postsContact)
const response = await tsRestClient.posts.update({
  params: { postId },
  body,
});

if (response.status === 200) {
  return response.body;
}

throw new UnexpectedError("Failed to update post", {
  response,
});


```

##

gingermusketeer:
Resolves #220

This PR adds a `strict` mode to the client which causes the client to throw errors for any HTTP status returned by the server  which is not defined in the contract.

### DIFF
--- a/libs/ts-rest/core/src/lib/client.spec.ts
+++ b/libs/ts-rest/core/src/lib/client.spec.ts
@@ -692,28 +692,36 @@ describe('custom api', () => {
     );
   });
 
-  it('has correct types when throwOnUnexpectedResponse is configured', async () => {
+  it('has correct types when strict is configured', async () => {
     const client = initClient(router, {
       baseUrl: 'https://api.com',
       baseHeaders: {
         'X-Api-Key': 'foo',
       },
-      throwOnUnexpectedResponse: true,
+      strict: true,
     });
 
-    fetchMock.getOnce(
-      {
-        url: 'https://api.com/posts',
-        headers: {
-          'Content-Type': 'application/json',
-        },
-      },
-      { status: 200 }
-    );
+    fetchMock.getOnce({ url: 'https://api.com/posts' }, { status: 200 });
 
     const result = await client.posts.getPosts({});
 
     // @ts-expect-error 404 is not defined in the known responses
     result.status === 404;
+  });
+
+  it('throws an error when strict is configured and response is unknown', async () => {
+    const client = initClient(router, {
+      baseUrl: 'https://isolated.com',
+      baseHeaders: {
+        'X-Api-Key': 'foo',
+      },
+      strict: true,
+    });
+
+    fetchMock.getOnce({ url: 'https://isolated.com/posts' }, { status: 419 });
+
+    await expect(client.posts.getPosts({})).rejects.toThrowError(
+      'Server returned unexpected response. Expected one of: 200 got: 419'
+    );
   });
 });
--- a/libs/ts-rest/core/src/lib/client.spec.ts
+++ b/libs/ts-rest/core/src/lib/client.spec.ts
@@ -691,4 +691,29 @@ describe('custom api', () => {
       })
     );
   });
+
+  it('has correct types when throwOnUnexpectedResponse is configured', async () => {
+    const client = initClient(router, {
+      baseUrl: 'https://api.com',
+      baseHeaders: {
+        'X-Api-Key': 'foo',
+      },
+      throwOnUnexpectedResponse: true,
+    });
+
+    fetchMock.getOnce(
+      {
+        url: 'https://api.com/posts',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+      { status: 200 }
+    );
+
+    const result = await client.posts.getPosts({});
+
+    // @ts-expect-error 404 is not defined in the known responses
+    result.status === 404;
+  });
 });
--- a/libs/ts-rest/core/src/lib/client.spec.ts
+++ b/libs/ts-rest/core/src/lib/client.spec.ts
@@ -692,13 +692,13 @@ describe('custom api', () => {
     );
   });
 
-  it('has correct types when strict is configured', async () => {
+  it('has correct types when throwOnUnknownStatus is configured', async () => {
     const client = initClient(router, {
       baseUrl: 'https://api.com',
       baseHeaders: {
         'X-Api-Key': 'foo',
       },
-      strict: true,
+      throwOnUnknownStatus: true,
     });
 
     fetchMock.getOnce({ url: 'https://api.com/posts' }, { status: 200 });
@@ -709,13 +709,13 @@ describe('custom api', () => {
     result.status === 404;
   });
 
-  it('throws an error when strict is configured and response is unknown', async () => {
+  it('throws an error when throwOnUnknownStatus is configured and response is unknown', async () => {
     const client = initClient(router, {
       baseUrl: 'https://isolated.com',
       baseHeaders: {
         'X-Api-Key': 'foo',
       },
-      strict: true,
+      throwOnUnknownStatus: true,
     });
 
     fetchMock.getOnce({ url: 'https://isolated.com/posts' }, { status: 419 });
--- a/libs/ts-rest/core/src/lib/client.spec.ts
+++ b/libs/ts-rest/core/src/lib/client.spec.ts
@@ -705,8 +705,9 @@ describe('custom api', () => {
 
     const result = await client.posts.getPosts({});
 
-    // @ts-expect-error 404 is not defined in the known responses
-    result.status === 404;
+    type ClientGetPostsResponseStatusType = Expect<
+      Equal<typeof result.status, 200>
+    >;
   });
 
   it('throws an error when throwOnUnknownStatus is configured and response is unknown', async () => {
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -177,11 +177,6 @@ export interface ClientArgs {
   api?: ApiFetcher;
   credentials?: RequestCredentials;
   jsonQuery?: boolean;
-  /**
-   * Ensures that the responses from the server match those defined in the
-   * contract.
-   */
-  throwOnUnknownStatus?: boolean;
 }
 
 export type ApiFetcherArgs = {
@@ -329,7 +324,7 @@ export const getCompleteUrl = (
 
 export const getRouteQuery = <TAppRoute extends AppRoute>(
   route: TAppRoute,
-  clientArgs: ClientArgs
+  clientArgs: InitClientArgs
 ) => {
   const knownResponseStatuses = Object.keys(route.responses);
   return async (inputArgs?: DataReturnArgsBase<any, ClientArgs>) => {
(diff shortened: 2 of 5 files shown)


libs/ts-rest/core/src/lib/client.spec.ts
```diff
@@ -691,4 +691,37 @@ describe('custom api', () => {
...
+      },
+      strict: true,
+    });
+
+    fetchMock.getOnce({ url: 'https://api.com/posts' }, { status: 200 });
+
+    const result = await client.posts.getPosts({});
+
+    // @ts-expect-error 404 is not defined in the known responses
+    result.status === 404;
```
Gabrola, line 709: I think you forgot this here

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -96,13 +107,16 @@ type DataReturnArgs<
   Without<DataReturnArgsBase<TRoute, TClientArgs>, never>
 >;
 
-export type ApiRouteResponse<T> =
+export type ApiRouteResponseStrict<T> =
```
michaelangrivera, line 110: Can you make a changeset explaining this change? Since this is a behavior change (probably welcomed), it'll be helpful for users to know that these response types are now strict. 

Also, maybe there's a way we can provide a "loose" type? Not sure if people would want that, though.

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -307,7 +341,7 @@ export const getRouteQuery = <TAppRoute extends AppRoute>(
       !!clientArgs.jsonQuery
     );
 
-    return await fetchApi({
+    const request = fetchApi({
```
Gabrola, line 344: Cleaner to `await` here.

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -319,6 +353,20 @@ export const getRouteQuery = <TAppRoute extends AppRoute>(
...
+
+    if (!clientArgs.strict) {
+      return await request;
+    }
+    const response = await request;
+
+    if (knownResponseStatuses.includes(response.status.toString())) {
+      return response;
+    }
+    throw new Error(
```
Gabrola, line 365: Would be helpful to throw a custom exception here exposing the response object through a property

Gabrola (changes requested):
Thank you for the contribution! The code looks great!

Although, we have a slightly different vision for how we need to implement type-safe "strict" status codes across both client and server, especially on the server to enforce it returning only defined status codes and that is what was discussed previously on the discord server. I still do see some merit in introducing this change specifically on the "vanilla" client.

In the react query client the data only includes types for defined 2xx statuses, and the "unknown" type is only exposed in the error state. So you could think of having kind of an escape hatch when coming across undefined "error" status codes. However, it will still fail if it comes across an undefined 2xx status since it will be handled as if it were whatever other 2xx status was actually in the contract. So it might actually make sense to introduce stricter handling in that library too.

The thing is, that even if you throw an exception when coming across an undefined status, you are still not guaranteed 100% error-free code since the body could be malformed and it would "break" through the type-safe escape hatch and cause some different kind of runtime error down the line.

There's also the case of error statuses that you are not in control of like 5xx errors or non-backend errors like gateway errors where the backend is incapable of forming a response, so I think it's just the nature of client-side code that there will always be the possibility of receiving an unexpected status code.

Anyway, I digress... since I think this could still be useful as an escape hatch and narrowing down the type, I think we should keep this change but rename the client option into something like `throwOnUnknownStatus`, as to not clash with our actual planned strict mode.

libs/ts-rest/core/src/lib/client.spec.ts
gingermusketeer: This is a type level assertion that ensures the flag works as expected. If this statement no longer causes a type error it will cause an error in the type check.

Would you prefer not to have this?

libs/ts-rest/core/src/lib/client.ts
gingermusketeer: Will do, wanted to get some feedback on the direction first to see if this was of interest.

libs/ts-rest/core/src/lib/client.ts
gingermusketeer: Agreed! Would you be open to having the exception configurable? I use an error library that captures additional information using async hooks and it would be helpful to be able to use one of these.

gingermusketeer:
@Gabrola Thanks for the feedback.

Implementing this on the server side as well makes sense. Perhaps it makes sense to tackle that in a separate PR?

Other than this what did you have in mind to achieve a strict implementation on both client and server?

libs/ts-rest/core/src/lib/client.spec.ts
gingermusketeer: Can see it in action here https://github.com/ts-rest/ts-rest/actions/runs/5105828907/jobs/9177654229?pr=228#step:12:79

Gabrola:
> @Gabrola Thanks for the feedback.
> 
> Implementing this on the server side as well makes sense. Perhaps it makes sense to tackle that in a separate PR?
> 
> Other than this what did you have in mind to achieve a strict implementation on both client and server?

Yeah, it definitely makes sense to have that in a separate PR!

I was planning to have the strict mode flag as part of the contract itself which would automatically remove the "unknown" response types on both the client and server TS types, but essentially do nothing in terms of runtime code. This fits our philosophy of being relatively as unobtrusive as possible

libs/ts-rest/core/src/lib/client.spec.ts
Gabrola: Ohh! That's nifty! I think that pattern could be useful in some cases. In this case, I think what we generally use to assert types would be more comprehensive here and also indicate the intention of the test clearly.

```typescript
Expect<Equal<typeof result.status, 200>>
```

libs/ts-rest/core/src/lib/client.ts
Gabrola: I think that might be a too specific feature, that could otherwise be covered by the `api` handler option, where you can catch the exception and re-throw your custom error.

```typescript
const client = initQueryClient(postsApi, {
  ...
  api: async (args) => {
    try {
      return tsRestFetchApi(args);
    } catch (error) {
      throw CustomException(error);
    }
  },
});
```

gingermusketeer: I wondered about that option but unless I am mistaken I don't think this would work as show above.

`getRouteQuery` is where this logic is implemented and it calls `fetchApi` which uses the custom api implementation.


Something like the following could work I think
```ts
const client = initQueryClient(postsApi, {
  ...
  api: async (args) => {
      const response = await tsRestFetchApi(args);
      if (!responseStatusExpected(response)) {
        throw CustomException("wrong status code")
      }
      return response
  },
});
```

But the hard part is knowing how to implement `responseStatusExpected` as I don't think the context is known at this point. The expected statuses could be passed down into the `api` function?

gingermusketeer:
> I was planning to have the strict mode flag as part of the contract itself which would automatically remove the "unknown" response types on both the client and server TS types, but essentially do nothing in terms of runtime code. This fits our philosophy of being relatively as unobtrusive as possible

This makes sense to me. Are you thinking the API would be something like one of these: 
```ts
// Option A
const c = initContract({strict: true})

// Option B
  posts: postsContract,
}, {strict:true})
```

I had a look at the server side of this change and I think this would make the changes needed there much easier.

libs/ts-rest/core/src/lib/client.ts
Gabrola: @gingermusketeer Oh you're right! It could be a simple change though, just need pass the `AppRoute` object along with the rest of the args. I think I can include this as part of a PR I am currently working on.

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -140,12 +155,33 @@ export type AppRouteFunction<
...
   baseUrl: string;
   baseHeaders: Record<string, string>;
   api?: ApiFetcher;
   credentials?: RequestCredentials;
   jsonQuery?: boolean;
+  /**
+   * Ensures that the responses from the server match those defined in the
+   * contract.
+   */
+  throwOnUnknownStatus?: boolean;
```
Gabrola, line 184: This should be removed from this interface as it is being used by the rest of the libraries

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -327,10 +373,17 @@ export type InitClientReturn<
   TClientArgs extends ClientArgs
 > = RecursiveProxyObj<T, TClientArgs>;
 
+export type InitClientReturnNoUnknownStatus<
+  T extends AppRouter,
+  TClientArgs extends ClientArgs & { throwOnUnknownStatus: true }
+> = RecursiveProxyObjNoUnknownStatus<T, TClientArgs>;
+
 export const initClient = <T extends AppRouter, TClientArgs extends ClientArgs>(
```
Gabrola, line 381: Regarding the other comment 

> This should be removed from this interface as it is being used by the rest of the libraries

```suggestion
export const initClient = <T extends AppRouter, TClientArgs extends ClientArgs & { throwOnUnknownStatus?: boolean; }>(
```
This should live here instead as it'll only be restricted to the vanilla client

Gabrola (changes requested):

libs/ts-rest/core/src/lib/unknown-status-error.ts
```diff
@@ -0,0 +1,15 @@
+export class UnknownStatusError extends Error {
```
Gabrola, line 1: Going to need to export this in index.ts so it's publicly accessible.

Sorry for the staggered reviews! 😅

gingermusketeer: No worries. Loving how detailed your reviews are!

libs/ts-rest/core/src/lib/client.ts
gingermusketeer: Done this. Pulled it out into a separate type so it could be reused in the two places needed.

Gabrola:
> > I was planning to have the strict mode flag as part of the contract itself which would automatically remove the "unknown" response types on both the client and server TS types, but essentially do nothing in terms of runtime code. This fits our philosophy of being relatively as unobtrusive as possible
> 
> This makes sense to me. Are you thinking the API would be something like one of these:
> 
> ```ts
> // Option A
> const c = initContract({strict: true})
> 
> // Option B
> export const contract = c.router({
>   posts: postsContract,
> }, {strict:true})
> ```
> 
> I had a look at the server side of this change and I think this would make the changes needed there much easier.

Yeah, it would be "Option B" in addition to allowing granular control on an endpoint-by-endpoint basis. We've introduced a pattern to do this kind control with the new headers feature, so it should be relatively simple.

libs/ts-rest/core/src/lib/client.ts
Gabrola: Perfect!

Gabrola (approved):
Great contribution! Thank you!

gingermusketeer:
Thank you for all the great feedback!

PR #231, opened 2023-05-28
feat: implement tanstack query cancellation

### LINKED ISSUE #222: Consume `AbortSignal` from react query and forward to API fetcher

Gabrola:
Resolves #222 

Screenshot 2023-05-29 013349

### DIFF
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -149,6 +149,7 @@ export interface ClientArgs {
 }
 
 export type ApiFetcherArgs = {
+  route: AppRoute;
   path: string;
   method: string;
   headers: Record<string, string>;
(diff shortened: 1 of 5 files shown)


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -188,6 +169,7 @@ export type ApiFetcherArgs = {
   rawQuery: unknown;
   contentType: AppRouteMutation['contentType'];
   credentials?: RequestCredentials;
+  signal?: AbortSignal;
```
michaelangrivera, line 172: Hmmm, is there a way to make this only part of the `initQueryClient` in the `@ts-rest/react-query` package? 
To me, it doesn't make sense to add this property/typing for the core package, since this is react-query specific?

Gabrola: It's actually a standard web API, you can use an AbortController to cancel any fetch request https://developer.mozilla.org/en-US/docs/Web/API/AbortController#examples

So it may be useful if you want to cancel large downloads/uploads, etc.

michaelangrivera: Oh sweet! ship it!

michaelangeloio (approved):

PR #232, opened 2023-05-29
feat(express): add context and per-endpoint middleware support

Gabrola:


apps/example-express/src/authenticated-app.ts
```diff
@@ -0,0 +1,185 @@
...
+    return {
+      status: 200,
+      body: params,
+    };
+  },
+});
+
+createExpressEndpoints(apiBlog, completedRouter, app, {
+  requestValidationErrorHandler: 'combined',
+  globalMiddleware: [
```
ulfdahlstrand, line 150: the same goes for this! ⭐

ulfdahlstrand (approved):

libs/ts-rest/express/src/lib/request-validation-error.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from 'zod';
+
+export class RequestValidationError extends Error {
```
oliverbutler, line 3: Sick. Love this ill add it to fastify once merged

oliverbutler (approved):
This looks awesome

libs/ts-rest/express/src/lib/get-value.ts
```diff
@@ -0,0 +1,21 @@
...
+  const value = path
+    .split(/[.[\]]/)
+    .filter(Boolean)
+    .reduce<GetFieldType<TData, TPath>>(
+      (value, key) => (value as any)?.[key],
+      data as any
+    );
+
+  return value !== undefined ? value : (defaultValue as TDefault);
+}
```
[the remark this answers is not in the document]
Gabrola: Yeah, this actually isn't being used at all, but it was exported though so I just put it in a separate file out of _(extreme)_ abundance of caution but out of sight lol. I have no idea why anyone would be using this though.

PR #233, opened 2023-05-29
Implement strict mode

gingermusketeer:
Implements a more generic strict mode than what was implemented in #228. This applies to both server and client side based on contract configuration rather than client configuration.


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -108,27 +108,27 @@ type DataReturnArgs<
...
-    }[keyof T];
+    }[keyof T['responses']];
 
-export type ApiRouteResponse<T> =
-  | ApiRouteResponseNoUnknownStatus<T>
-  | {
-      status: Exclude<HTTPStatusCode, keyof T>;
-      body: unknown;
-    };
+export type ApiRouteResponse<T extends AppRoute> = T extends { strict: true }
```
gingermusketeer, line 119: Is this breaking change acceptable?

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -352,7 +350,8 @@ export const getRouteQuery = <TAppRoute extends AppRoute>(
       },
     });
 
-    if (!clientArgs.throwOnUnknownStatus) {
+    const strictMode = route.strict ?? clientArgs.throwOnUnknownStatus;
```
gingermusketeer, line 353: Perhaps `throwOnUnknownStatus` can be removed in a follow up PR?

libs/ts-rest/express/src/lib/ts-rest-express.spec.ts
```diff
@@ -1,4 +1,51 @@
...
+    });
+  });
+
+  it('does not allow unknown statuses when in strict mode', () => {
+    const cStrict = c.router({ posts: postsRouter }, { strict: true });
+    const s = initServer();
+
+    s.router(cStrict, {
+      posts: {
+        // @ts-expect-error 201 is not defined as a known response
```
gingermusketeer, line 38: I wanted to test all the types combined so used this approach. Let me know if you would prefer the `Expect<Equal<>>` approach instead.

libs/ts-rest/core/src/lib/client.ts
Gabrola: Unfortunately not, however, we can instead of move the conditional type logic to the `ApiResponseForRoute` type alias.

libs/ts-rest/core/src/lib/dsl.ts
```diff
@@ -23,6 +23,7 @@ export type AppRouteQuery = {
   description?: string;
   deprecated?: boolean;
   responses: Record<number, unknown>;
+  strict?: boolean;
```
Gabrola, line 26: Can this be a more descriptive name, maybe `strictResponses`?

libs/ts-rest/core/src/lib/client.ts
Gabrola: Since the default behavior is not throw on any response, it might catch some people off guard if we start throwing when strict mode is enabled. I think we should only throw on `throwOnUnknownStatus` and not strict mode. 

I know it's technically a non-breaking change, but just changing an option in the contract is not immediately obvious that it might affect the runtime behavior of the client, so we should keep strict mode for type-safety only for now.

Gabrola (changes requested):
This is great, thank you! Left you a couple of comments, I'm slightly tired to do an in-depth review, so maybe later but it looks good on initial glance 👌

libs/ts-rest/core/src/lib/dsl.ts
Gabrola: Actually that's slightly ambiguous too because it might be mistaken to do the same functionality of `validateResponses`. Maybe `strictStatusCodes`?

libs/ts-rest/core/src/lib/client.ts
gingermusketeer: Will do that!

libs/ts-rest/core/src/lib/client.ts
gingermusketeer: Perhaps some type docs would help in this case? That way they get information in their text editor if it is setup to see the method docs. 

Something that your comment made me wonder was what we should do on the server side if there is an invalid response. Should it also support this?

If you want to go down this path then perhaps a strict mode option would be better? For example `strict: throwOnUnknownStatus` and this would apply on both sides?

Gabrola: Well yeah the main point from this strict mode is to narrow down the types to exclude all undefined status codes and the corresponding `unknown` body type from the responses. So you should get a type error if you try to return a response from the server with a status that is not defined in the contract. So the complement to that is to only narrow the type on the client side as well and do nothing else.

ts-rest's philosophy is to be as light and unobtrusive as possible, so we should target to enforce this only through type-safety by default. Any extra "guardrails" that a developer might want should be optional and explicitly enabled rather than enforcing it by default or even not having the option to turn it off.

gingermusketeer: This makes sense. Thanks for that context.

With this in mind I suggest two strict modes:
- `strict: 'responseTypes'`
- `strict: 'throwOnUnknownResponse'`

This allows the cognitive load for users to be kept to a minimum, due to consolidated configuration, while allowing them to choose the behaviour that they want. This would replace `throwOnUnknownStatus`. If in the future this causes any unforeseen issues in the future then there is a way to add additional modes. What do you think? If you don't like it then happy to go with `strict: true` and `throwOnUnknownStatus`.

Gabrola: Yeah that sounds very reasonable. I think there's also the point that the current change would only affect the vanilla client, but not the other clients tanstack query clients since they already have their own error handling, so it wouldn't really make sense to have an option in the contract itself that's only specific to one client, so to me it makes sense that the "throw" option should live in that client's options rather than in the contract.

libs/ts-rest/core/src/lib/dsl.ts
gingermusketeer: Changed it to `strictStatusCodes`

libs/ts-rest/core/src/lib/client.ts
gingermusketeer: Changed it to only throw if  `throwOnUnknownStatus` is set.

libs/ts-rest/core/src/lib/client.ts
gingermusketeer: Undone this change :)

libs/ts-rest/core/src/lib/dsl.ts
```diff
@@ -108,6 +110,9 @@ type ApplyOptions<
 > = Omit<TRoute, 'headers'> &
   WithoutUnknown<{
     headers: UniversalMerge<TOptions['baseHeaders'], TRoute['headers']>;
+    strictStatusCodes: TOptions['strictStatusCodes'] extends true
+      ? true
+      : TRoute['strictStatusCodes'];
```
Gabrola, line 115: I think the route specific option should take precedence regardless of it's true or false. This will allow a user to override the one in the options with whatever they like.
 
```suggestion
    strictStatusCodes: TRoute['strictStatusCodes'] extends boolean
      ? TRoute['strictStatusCodes']
      : TOptions['strictStatusCodes'] extends boolean
      ? TOptions['strictStatusCodes']
      : unknown;
```

The `unknown` type will cause the property to be stripped by the `WithoutUnknown` type.

libs/ts-rest/core/src/lib/dsl.ts
```diff
@@ -186,6 +192,8 @@ const recursivelyApplyOptions = <T extends AppRouter>(
           {
             ...value,
             headers: zodMerge(options?.baseHeaders, value.headers),
+            strictStatusCodes:
+              options?.strictStatusCodes ?? value.strictStatusCodes,
```
Gabrola, line 196: Same as the other comment
```suggestion
            strictStatusCodes:
              value.strictStatusCodes ?? options?.strictStatusCodes,
```

libs/ts-rest/express/src/lib/ts-rest-express.spec.ts
Gabrola: This approach makes sense here, so all good!

libs/ts-rest/core/src/lib/dsl.ts
gingermusketeer: Thanks have made this change!

gingermusketeer:
Thanks for all the help and the tidy up! Looks great :)

Gabrola:
@all-contributors please add @gingermusketeer for code and ideas

PR #235, opened 2023-05-29
Changeset fixes + GHA cleanup

### LINKED ISSUE #234: changeset action does not create github release
**Background**:
For some reason, the GHA https://github.com/changesets/action stopped creating github "releases". This section is useful for devs who are looking to see any latest changes (or breaking changes) from our changeset log. 

**Done**:
Investigate and resolve the issue

michaelangrivera:
Fixes #234


.github/workflows/release.yml
```diff
@@ -42,24 +56,19 @@ jobs:
...
 
       - name: Create Release Pull Request or Publish to npm
-        uses: changesets/action@master
+        uses: changesets/action@d89c1de63c7f28ac47ec85ed395f5f1d045d4697
         with:
           title: Release Tracking
           # this expects you to have a script called release which does a build for your packages and calls changeset publish
           publish: pnpm release
           version: pnpm version-packages
+          createGithubReleases: true
```
michaelangrivera, line 65: Just in case

.github/workflows/release.yml
```diff
@@ -9,14 +9,28 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       nx_affected: ${{ steps.set-nx-affected.outputs.nx_affected }}
+    permissions:
```
michaelangrivera, line 12: We need to add `permissions` for better supply chain security. We might need to experiment with this, but these permissions should be sufficient. Let me know if this needs to be changed!

Gabrola (approved):
Can't really tell if this is going to work without testing lol

PR #236, opened 2023-05-30
chore: use aggregate release for all packages

Gabrola:
Using the "dotansimha/changesets-action" fork of "changesets/action" which allows us to create a single release for all packages together. Should be better than creating a release per package. Separate tags per package will still be created. This might fix the problem with releases not being created. 

Also cleaned up anything that has to do with `nx_affected` in the release since that was only used previously to determine whether to release docs or not. Vercel now takes care of that, so it is not needed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,8 +7,6 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    outputs:
-      nx_affected: ${{ steps.set-nx-affected.outputs.nx_affected }}
     permissions:
       actions: read
       contents: write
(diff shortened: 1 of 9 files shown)


.github/workflows/release.yml
```diff
@@ -20,15 +18,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set NX Shas
-        uses: nrwl/nx-set-shas@177b48373c6dc583ce0d9257ffb484bdd232fedf
```
Gabrola, line 24: Was only needed to run the `ci:nx-affected` command

oliverbutler (approved):
Nice! I think this LGTM will be glad to get these back.

PR #237, opened 2023-05-30
Add Fastify Adapter 🎉

### LINKED ISSUE #217: How about Fastify support?
Now I can select backend

```
yarn add @ts-rest/nest @ts-rest/express
```

but I would like to type

```
yarn add @ts-rest/fastify
```

and use this awesome tool with https://www.fastify.io/

oliverbutler:
Closes #217

Please note this is a work in progress - would love to get feedback asap so we can iterate on this, I'm pretty new to Fastify (as of 3 hours ago starting this)

Mostly stole our express library, but refactored loads of parts to be much better.


libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
```diff
@@ -0,0 +1,237 @@
...
+  ) {
+    reply.status(400).send({
+      queryParameterErrors: queryResult.success ? null : queryResult.error,
+      pathParameterErrors: paramsResult.success ? null : paramsResult.error,
+      headerErrors: headersResult.success ? null : headersResult.error,
+      bodyErrors: bodyResult.success ? null : bodyResult.error,
+    });
+
+    return null;
+  }
```
Gabrola, line 111: This is great! This also made me think of how there is some need to customize these errors. What do you think that instead of setting the response here we instead throw a custom error `TsRestValidationError` that contains all of these error parameters.

We also provide a standard error handler to be used by the developer like this `fastify.setErrorHandler(TsRestErrorHandler)` that takes care returning the standard format you have here, but can be swapped out at will

libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
```diff
@@ -0,0 +1,237 @@
...
+      const validationResults = validateRequest(
+        request,
+        reply,
+        appRoute,
+        options
+      );
+
+      if (validationResults === null) {
+        return;
+      }
```
Gabrola, line 168: If we throw in `validateRequest` we won't need this check

libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
```diff
@@ -0,0 +1,237 @@
...
+  path: string[],
+  fastify: fastify.FastifyInstance,
+  options: RegisterRouterOptions
+) => {
+  if (typeof routerImpl === 'object') {
+    for (const key in routerImpl) {
+      recursivelyRegisterRouter(
+        routerImpl[key] as unknown as RecursiveRouterObj<T>,
+        appRouter,
+        [...path, key],
```
Gabrola, line 220: Would be easier, rather than having to propagate the path, if we instead pass the sub-routers in each iteration. So remove the `path` and just pass

```typescript
        routerImpl[key] as unknown as RecursiveRouterObj<T>,
        appRouter[key],
```

Will no longer need the `getValue` function as well

Gabrola:
This looks great! ❤

libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
oliverbutler: Good shout. I'll do the same in the express lib - unless you're doing that rewrite?

Gabrola: I was thinking of doing it but it'd be a breaking change because you'd need to manually mount the middleware unless I can figure out a clean way to mount the new middleware by default and allow a user to override that. Let me see.

Gabrola: @oliverbutler btw what I meant for this comment for to be combined with the approach in the comment above. I'm not sure what the behavior would be if you throw _anything_ but still return a response.

Anyway, you can check the approach I took here https://github.com/ts-rest/ts-rest/pull/232/commits/2d0ca77bbd8d2bee7e2a123bc5fdae6fc07ce89c

I followed the same format for the new `combined` error handler, but had to keep an old `default` one for backwards compatibility, but we won't need that problem for fastify. So in this case you can just either consume a custom error handler from the developer or use our own to return that 400 error with this new format.

apps/docs/docs/fastify.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+```typescript
+createfastifyEndpoints(contract, router, app, {
+  validateResponses: true,
+});
+```
+
+### Request Validation Error Handling
+
+The default functionality of handling request validation errors is to return a 400 response with the first validation error that the validator comes across.
```
Gabrola, line 90: I think this only applies to the express lib

Gabrola (approved):
🔥😍
🔥😍
🔥😍
🔥😍

apps/docs/docs/fastify.mdx
oliverbutler: Ah! Thanks 🪓

PR #251, opened 2023-06-03
Fix nest response shape typing

gingermusketeer:
Couple of changes following up on #233

- Passes through AppRoute to ApiRouteServerResponse type correctly
- Undoes change to NestResponseShapes which made return types a promise. This makes requires consumers to wrap the type in `Awaited` to be able to use it.

### DIFF
--- a/libs/ts-rest/nest/src/lib/ts-rest-nest.spec.ts
+++ b/libs/ts-rest/nest/src/lib/ts-rest-nest.spec.ts
@@ -63,3 +63,24 @@ it('does not allow unknown statuses when in strict mode', () => {
     }
   }
 });
+
+it('allows responseShapes types to be used in controller logic', () => {
+  const cStrict = c.router({ posts: postsRouter }, { strictStatusCodes: true });
+  const nestContract = nestControllerContract(cStrict.posts);
+  type contractType = typeof nestContract;
+  type RequestShapes = NestRequestShapes<contractType>;
+  type ResponseShapes = NestResponseShapes<typeof nestContract>;
+
+  class PostController implements NestControllerInterface<typeof nestContract> {
+    @TsRest(nestContract.getPost)
+    async getPost(
+      @TsRestRequest() { params: { id } }: RequestShapes['getPost']
+    ) {
+      const result: ResponseShapes['getPost'] = {
+        status: 200 as const,
+        body: null,
+      };
+      return result;
+    }
+  }
+});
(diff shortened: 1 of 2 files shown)


libs/ts-rest/nest/src/lib/ts-rest-nest.ts
```diff
@@ -31,11 +31,9 @@ type AppRouterRequestShapes<T extends AppRouter> = Without<
 type AppRouterResponseShapes<T extends AppRouter> = Without<
   {
     [K in keyof T]: T[K] extends AppRoute
-      ? Promise<
```
gingermusketeer, line 34: Having this wrapped in a promise means that you can't do:

```ts
const post: NestResponseShapes<typeof contract>['getPost'] = {...}
```

instead need to do:

```ts
const post: Awaited<NestResponseShapes<typeof contract>['getPost']> = {...}
```

libs/ts-rest/nest/src/lib/ts-rest-nest.spec.ts
```diff
@@ -22,7 +23,12 @@ it('allows unknown statuses when not in strict mode', () => {
   const cLoose = c.router({ posts: postsRouter });
   const nestContract = nestControllerContract(cLoose.posts);
   type RequestShapes = NestRequestShapes<typeof nestContract>;
+  type ResponseShapes = NestResponseShapes<typeof nestContract>;
 
+  const responseTypeCheck: Awaited<ResponseShapes['getPost']> = {
```
gingermusketeer, line 28: Would be nicer to use `Expect<Equal<>>` here but I don't think it is in scope as it is in the core library?

Gabrola:
Ah nice one! Thank you!

libs/ts-rest/nest/src/lib/ts-rest-nest.ts
Gabrola: > Undoes change to NestResponseShapes which made return types a promise.

So what I understand from that is that `AppRouterResponseShapes` was not wrapped in a Promise and then was changed to be a Promise type. When was that "change" introduced?

If it was long ago, it's unfortunate, as the change back might break some users' code who are doing something like this
```typescript
async getPosts(
    @TsRestRequest()
    _: RequestShapes['getPosts']
  ): ResponseShapes['getPosts'] { ... }
```

Since the function would be expected a non-promise type but the async function is returning a promised type.

Gabrola: Ah, I just checked and it was in the last PR. Let's ship it then!

Gabrola (approved):
LGTM! Thank you for the quick follow up! 🎉

gingermusketeer:
Thanks for the fast merge!

PR #256, opened 2023-06-03
Fix cjs usage of libraries

gingermusketeer:
Trying to upgrade from 3.22 to 3.25 I am getting the following error:

```
    const exports = origRequire.apply(this, arguments)
                                ^
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/maxbrosnahan/project/node_modules/@ts-rest/nest/index.js from /Users/maxbrosnahan/project/dist/main.js not supported.
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/maxbrosnahan/project/node_modules/@ts-rest/nest/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).

```

To fix this I have added an override in the fix-esm script.

## Before
Skjermbilde 2023-06-03 kl  22 02 11

## After
Skjermbilde 2023-06-03 kl  22 06 53

### DIFF
--- a/.changeset/moody-falcons-confess.md
+++ b/.changeset/moody-falcons-confess.md
@@ -1,0 +1,12 @@
+---
+'@ts-rest/core': patch
+'@ts-rest/express': patch
+'@ts-rest/fastify': patch
+'@ts-rest/nest': patch
+'@ts-rest/next': patch
+'@ts-rest/open-api': patch
+'@ts-rest/react-query': patch
+'@ts-rest/solid-query': patch
+---
+
+Fix ESM/CJS issues in package.json
--- a/libs/ts-rest/core/project.json
+++ b/libs/ts-rest/core/project.json
@@ -27,8 +27,7 @@
         "format": ["esm", "cjs"],
         "compiler": "tsc",
         "rollupConfig": "tools/scripts/rollup.config.js",
-        "generateExportsField": true,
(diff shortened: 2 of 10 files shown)


Gabrola (approved):
Ooof, looks like the upgrade to nx 16 rendered "skipTypeField" useless. Nice catch!

Gabrola:
Thank you @gingermusketeer! You've been helping greatly so far! 🙏

oliverbutler:
@Gabrola this is probs our second or third issue caused (in part) because of NX weirdness with building 

I wonder if we should have a test which installs and builds ts rest NOT in NX so we can confirm the dist copies work well. Wdyt?

Gabrola:
> @Gabrola this is probs our second or third issue caused (in part) because of NX weirdness with building
> 
> I wonder if we should have a test which installs and builds ts rest NOT in NX so we can confirm the dist copies work well. Wdyt?

Yeah that's a good idea. Some very basic smoke testing would have caught this.

gingermusketeer:
Thanks for the merge ❤️  I have upgraded and tested the changes and it is working :)

PR #258, opened 2023-06-03
Added recursive `pathPrefix` in contract options

### LINKED ISSUE #215: Prefixed URL
Hello guys, there is a way to use prefixed url or nested calls in the router, like this?

```
class App {
  public app: express.Application;

  constructor() {
    this.app = express();
    this.config();
    this.routerSetup();
    this.errorHandler();
  }

  private config() {
    // view engine setup
    this.app.set('views', path.join(__dirname, 'views'));
    this.app.set('view engine', 'jade');

    this.app.use(logger('dev'));
    this.app.use(express.json());
    this.app.use(express.urlencoded({ extended: false }));
    this.app.use(cookieParser());
    this.app.use(express.static(path.join(__dirname, 'public')));
  }

  private routerSetup() {
    this.app.use('/', indexRouter);
    this.app.use('/users', usersRouter);
    this.app.use(V1(this.app, '/v1'));
  }

```

```

const c = initContract();

export const postContract = c.router({
  createPost: {
    method: 'POST',
    path: '/posts',
    responses: {
      201: c.response<Post>(),
    },
    body: c.body<{title: string}>(),
    summary: 'Create a post',
  },
  getPost: {
    method: 'GET',
    path: `/posts/:id`,
    responses: {
      200: c.response<Post | null>(),
    },
    summary: 'Get a post by id',
  },
});

```

```
export function V1(app: Application, url: string): Router {
  return Router().use('/test', PostHandler.register());
}
```

```
    static register() {
        return Router()
            .use(
                '/groups/:groupId/post',
                GroupHandler.register(connection)
            )
            .use('/groups/:groupId/post/:id/comments', GroupCommentHandler.register())
            .use(
                '/groups/:groupId/post/:id',
                GroupHandler.register()
            )}
```

I want those urls in that file to be  like 'v1/test/groups/:groupId/post' without having to rewrite v1/test every time.
Because some times I cannot be sure that will be v1/test, since I'm creating a larger api with many modules.
Maybe make createExpressEndpoints reusable to make a chain of "router.use".

Jwcode-uk:
closes #215

Added changes to allow user to supply path prefix in options and have it added to path recursively.


libs/ts-rest/core/src/lib/dsl.spec.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import { z } from 'zod';
+import { TypeOf, z } from 'zod';
```
oliverbutler, line 2: Can we remove this

libs/example-microservice/util-posts-api/src/lib/example-microservice-util-posts-api.ts
```diff
@@ -19,35 +19,38 @@ export type Post = z.infer<typeof PostSchema>;
...
+          message: z.string(),
+        }),
+      },
+      description: 'Update post thumbnail',
     },
-    description: 'Update post thumbnail',
   },
-});
+  { urlPrefix: 'test' }
+);
```
oliverbutler, line 56: Why did we change this?

oliverbutler:
Damn - that was quick 🎉 Great to see this was q quick win

cat-speed

libs/example-microservice/util-posts-api/src/lib/example-microservice-util-posts-api.ts
Jwcode-uk: No Comment 👀

libs/ts-rest/core/src/lib/dsl.spec.ts
Jwcode-uk: sorted :)

oliverbutler (approved):
lgtm

libs/ts-rest/core/src/lib/dsl.ts
```diff
@@ -128,9 +131,10 @@ export type AppRouter = {
   [key: string]: AppRouter | AppRoute;
 };
 
-export type RouterOptions = {
+export type RouterOptions<TPrefix extends string> = {
```
Gabrola, line 134: I think we can default this to `''` so we don't have to explicitly pass `<string>` everywhere.

```suggestion
export type RouterOptions<TPrefix extends string = ''> = {
```

libs/ts-rest/core/src/lib/dsl.spec.ts
```diff
@@ -477,4 +477,27 @@ describe('contract', () => {
       >
     >;
   });
+
+  describe('pathPrefix', () => {
+    it('Should recursively apply pathPrefix to path', () => {
```
Gabrola, line 482: Might be useful to add a type assertion test just like the test above to make sure the runtime path and TS type are the same

Gabrola:
Wow that was quick! Super impressed how you navigated around this one so quickly!

libs/ts-rest/core/src/lib/dsl.ts
Gabrola: Actually I might be slightly confused, it's late lol but I think the correct way would be `TPrefix extends string = string`

Gabrola: @Jwcode-uk I pulled your code locally just to make sure I wasn't just spewing fairy dust, and it indeed worked, so I pushed the changes. I hope you don't mind! 🙏

Gabrola (approved):

libs/ts-rest/core/src/lib/dsl.ts
Jwcode-uk: Appreciate the help!!

PR #260, opened 2023-06-03
fix: broken query encoding for keys or values containing equals characters

Gabrola:


oliverbutler:
@Gabrola damn good catch. Much better implementation 😅 

Did you find this on your codebase or where was this reported?


moira

oliverbutler (approved):
LGTM :D

Gabrola:
> @Gabrola damn good catch. Much better implementation 😅
> 
> Did you find this on your codebase or where was this reported?
> 
> moira

@oliverbutler I was working on the super not top-secret serverless lib and needed to create the inverse function of `encodeQueryParams` and I thought something didn't look right when going through the code and the unit test confirmed it lol

PR #263, opened 2023-06-04
feat: allow defining non-json responses in the contract

### LINKED ISSUE #180: Capability to send non-JSON responses
The express and next server-side implementations do not allow the request handlers to return anything other than JSON since we are returning the response using `res.status(statusCode).json(body)`

There is no way to prevent this behavior so we'd need to expose the response object to the request handler functions, in addition, we need to define a method to prevent the default behavior of calling `.json` on the response object.

Might need to re-visit how we can define these responses in the contract if possible.

This problem doesn't exist in the NestJS implementation since we do a `res.status(...); return body;` in the interceptor, and the handler can access the response object by injecting it using the `@Res` decorator. Will need to double check that calling `res.status(...);` does not cause problems in case the status and/or response body was already sent manually by the handler.

Gabrola:
Resolves #180


oliverbutler:
My other question - should we even have a different response type for this rather than just a "bodyContentType": "text/html" | "text/...

My reasoning

Unable to use a Zod object for the response with your way
You loose body validation (which can be done against text/plain by zod happily)
Less naming to decide 😄

```ts
getIndex: {
        method: 'GET',
        path: `/index.html`,
        responses: {
          200: {
            contentType: "text/html",
            body: c.body<string>()
          }
        },
      },
      getRobots: {
        method: 'GET',
        path: `/robots.txt`,
        responses: {
          200: {
            contentType: "text/plain",
            body: z.string().refine((str) => str === ":)")
          }
        },
      },
      getCss: {
        method: 'GET',
        path: '/style.css',
        responses: {
          200: {
            contentType: "text/plain",
            body: ValidCssSchema
          },
      },
```

This feels more consistent with our design philosphy no?
Also then you get type info for OpenAPI a little easier

```ts
responses: {
   200: MyReturnSchema // defaults to application/json as a super sane default,
   400: {
      contentType: "text/plain",
      body: z.string()
   }
}
```

libs/ts-rest/core/src/lib/dsl.ts
```diff
@@ -173,13 +190,27 @@ type ContractInstance = {
...
+  /**
+   * Define a custom response type
+   */
+  otherResponse: <T extends ContractAnyType>({
+    contentType,
+    body,
+  }: {
+    contentType: string;
+    body: T;
+  }) => ContractOtherResponse<T>;
```
oliverbutler, line 213: We're gonna see so many warnings in our prod code 🙈

oliverbutler (approved):
This looks great!! Much improved over what I was doing the other day 🙌

libs/ts-rest/core/src/lib/dsl.ts
Gabrola: 😆

PR #252, opened 2023-06-03
feat: add fastify plugin implementation

baryla:
This PR adds the Fastify plugin implementation which aims to replace `registerRouter` in which we have to pass more arguments than we actually need. 

This does however change the return type of the `router` method on `initServer` but there shouldn't be any unintended consequences from this change.

PR #265, opened 2023-06-04
feat: add middleware support to fastify

baryla:
This PR adds support for middleware in the fastify implementation. This is heavily inspired by the express implementation so the code should look somewhat familiar :) 

First-class middleware support was removed from Fastify after v3 and now relies on plugins like `@fastify/express` `@fastify/middie` etc.

If this PR is good to be merged, I can also update the docs to reflect these changes too.

One major point is that the middleware can only be used if the router is installed as a `plugin` (#252) because the middleware support is also plugin based and in order to get access to the plugin context, we also need to be in the plugin context.


oliverbutler:
This looks great - nice and aligned with the new express middleware (bar global, assuming because you can write it outside of ts-rest?)

@Gabrola would be great to hear your thoughts also 🙌

libs/ts-rest/fastify/src/lib/ts-rest-fastify.spec.ts
```diff
@@ -287,4 +288,220 @@ describe('ts-rest-fastify', () => {
...
+    });
+
+    app.register(s.plugin(router));
+
+    await app.ready();
+
+    const response = await supertest(app.server).get('/middleware').send();
+
+    expect(response.statusCode).toEqual(401);
+  });
```
oliverbutler, line 506: What happens if a user **doesnt** do `app.register(s.plugin(router));` and they use the `registerRoute` instead?

Is an error thrown? Can we catch it with TS? 🤔

libs/ts-rest/fastify/package.json
```diff
@@ -2,10 +2,18 @@
...
     "zod": "^3.0.0"
   },
   "peerDependenciesMeta": {
+    "@fastify/express": {
+      "optional": true
+    },
+    "@fastify/middie": {
+      "optional": true
+    },
     "zod": {
```
oliverbutler, line 17: Are they required for this? We only use these in the spec right?

Gabrola:
I think Fastify hooks have replaced the middleware functionality without needing any plugins and can still be added either globally or per-route, with the added benefit of the granular hook types (onRequest, preHandler, etc.).

Also from [this page](https://www.fastify.io/docs/latest/Reference/Middleware/) there seems to be a major limitation regarding the use of middleware:

> Fastify middleware does not expose the send method or other methods specific to the Fastify [Reply](https://www.fastify.io/docs/latest/Reference/Reply#reply) instance. This is because Fastify wraps the incoming req and res Node instances using the [Request](https://www.fastify.io/docs/latest/Reference/Request#request) and [Reply](https://www.fastify.io/docs/latest/Reference/Reply#reply) objects internally, but this is done after the middleware phase. If you need to create middleware, you have to use the Node req and res instances. Otherwise, you can use the preHandler hook that already has the [Request](https://www.fastify.io/docs/latest/Reference/Request#request) and [Reply](https://www.fastify.io/docs/latest/Reference/Reply#reply) Fastify instances. For more information, see [Hooks](https://www.fastify.io/docs/latest/Reference/Hooks#hooks).

I'm not super familiar with the fastify ecosystem but some quick research indicates that popular packages like `@fastify/passport` depend on hooks rather than middleware, and they cannot be used interchangeably because of the different exposed objects as quoted above. So to support these packages we must add hook functionality.

Personally, I think its better to just leave behind an officially unsupported system behind while not needing to rely on extra external packages, and it seems like it's a relatively easy switch from the perspective of this code.

baryla:
After reading some more on this, I think you're right @Gabrola and the approach I took doesn't really work well with Fastify anymore. 

My main reason for going this way was to align with ts-rest's express API but I'm now sure that this isn't the right way. 

Taking into account fastify's hook system, what do we think about an API like this: 
```ts
  hooks: {
    preValidation: (request, reply) => {
      // single hook 
    },
    onRequest: [
      (request, reply) => {
        // array of hooks
      }
    ]
  },
  getPosts: {
    handler: async (tx) => {
      return {
        status: 200,
        body: []
      }
    }
  }
});
```

This way, we can support most of the existing fastify plugins, like the Passport one that @Gabrola mentioned. 

All of the hooks' `request` object will also have the `tsRestRoute` like the express one to keep consistency. 

Thoughts? @oliverbutler

oliverbutler:
Can the passport plug-in be used without this api? Just wondering as this feels just like we've got a api for running code, just before other code - which just sounds like an early return with extra steps no? 

You guys are the expects here, more than happy to take your ways.

( @baryla worry about grand fathering in our api design, other than the Nest lib we need to give some love to our APIs)

baryla:
The hook examples I gave would be direct mappings to what fastify supports. They have a bunch of request hooks and 2 of them are `preValidation` and `onRequest` so it wouldn't necessarily be an early return but actually what the user would provide here would be "piped" into the request instance. 

What this would in theory look like, with the passport plugin;

```ts
import fastifyPassport from '@fastify/passport';

const app = fastify({ logger: false });
app.register(fastifyExpress);

server.register(fastifyPassport.initialize());

  getPosts: {
    hooks: {
      preValidation: [
        fastifyPassport.use(new SomeStrategy(...), { failureRedirect: '/login' }),
        // .. more pre validation hooks here
      ]
    },
    handler: async (tx) => {
      return {
        status: 200,
        body: []
      }
    }
  }
});
```

Does that make sense?

Gabrola:
@baryla yup, that looks perfect! And I'd guess we'd have the a `globalHooks` option too for the entire router. Also this is probably going to bring the internal logic to be much more similar to the express implementation where we add the middleware directly with the route inside the `fastify.route({ ...hooks, handler: ... })` just like in express `app.get(path, ...handlers)`

Also on a quick glance it doesn't seems like we can even what we want with middleware, the way it's like in the code currently `app.use(...middleware).route(...)` seems to indicate that we are adding the route's middleware to the entire app not just specific routes. But based on docs [here](https://www.fastify.io/docs/latest/Reference/Middleware/#restrict-middleware-execution-to-certain-paths), you need to use `app.use(path, ...middleware)` to add middleware to specific paths and it does not support routes with parameters or wildcards. So what we're offering with express wouldn't be possible unless we hack up the middleware implementation even further to support our use-case with route-specific middleware.

baryla:
@oliverbutler @Gabrola what do you guys think about the changes from middlewares to hooks?

libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
```diff
@@ -127,7 +157,19 @@ export const initServer = () => ({
...
+      contract,
+      [],
+      app,
+      restOfOptions
+    );
+
+    Object.keys(hooks).forEach((hookName) =>
+      // @ts-expect-error Fastify's hook overload seems to be complaining here for no reason afaik
+      app.addHook(hookName, hooks[hookName])
+    );
```
Gabrola, line 172: This will add the hooks to the entire app rather than just the routes of the passed contract. We should forward the entire `options` object to `recursivelyRegisterRouter` and add these "global" hooks to each route individually right before the route's own hooks.

Gabrola (changes requested):
Sorry for the delay @baryla, I didn't notice that you've finished this 🙏

gustawdaniel:
It would be awesome to see the option to attach middleware at different stages.

In fastify we have

```
Request/Reply Hooks
- onRequest
- preParsing
- preValidation
- preHandler
- preSerialization
- onError
- onSend
- onResponse
- onTimeout
- onRequestAbort
- Manage Errors from a hook
- Respond to a request from a hook
Application Hooks
- onReady
- onClose
- preClose
- onRoute
- onRegister
```

 https://fastify.dev/docs/latest/Reference/Hooks/

Now I am trying to attach middleware `onError` but in docs, there is described only `preHandler` in the sentence

> These run before any route-specific middleware.

https://ts-rest.com/docs/express/middleware

michaelangelo-io:
hi @baryla , do you mind fixing the merge conflict?

oliverbutler:
Gonna close for now, its been a hot while since this was active

PR #267, opened 2023-06-04
Docs: added pathPrefix to docs. also added Options section.

Jwcode-uk:
added docs explaining pathPrefix.

Also added a section to keep all options docs in one place - if you disagree with this I can revert :)

### DIFF
--- a/apps/docs/docs/core/core.md
+++ b/apps/docs/docs/core/core.md
@@ -227,9 +227,9 @@ export const contract = c.router({
 
 ### Path Prefix
 
-The `pathPrefix` option allows you to add a prefix to URL paths, thereby promoting more modular and reusable routing logic. This feature is applied recursively, thereby facilitating the application of prefixes to nested contracts. In addition, when hovering over the contract, the prefixed path will appear at the beginning of the path.
+The `pathPrefix` option allows you to add a prefix to paths, allowing more modular and reusable routing logic. This feature is applied recursively, allowing the application of prefixes to nested contracts. In addition, when hovering over the contract, the prefixed path will appear at the beginning of the path for ease of use.
 
-Here is an example of how to utilize the `pathPrefix` option. In this scenario, the resulting path is `/api/v1/MyPath`.
+Here is an example of how to use the `pathPrefix` option. In this example, the resulting path is `/api/v1/MyPath`.
 
 ```typescript
 const c = initContract();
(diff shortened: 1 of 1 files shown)


apps/docs/docs/core/core.md
```diff
@@ -176,7 +176,24 @@ export const contract = c.router({
...
+You can assign `baseHeaders` which will be merged with the contract `headers`. Here's how to set it:
+
+```typescript
+const c = initContract();
+export const contract = c.router({
+  // ...endpoints
+}, {
+  strictStatusCodes: true,
+});
+```
```
oliverbutler, line 194: You said `heres how to set it` then didn't show how to set it :D

apps/docs/docs/core/core.md
```diff
@@ -207,3 +224,40 @@ export const contract = c.router({
...
+
+The `pathPrefix` option allows you to add a prefix to paths, allowing more modular and reusable routing logic. This option is applied recursively, allowing the application of prefixes to nested contracts. In addition, when hovering over the contract, the prefixed path will appear at the beginning of the path for ease of use.
+
+Here is an example of how to use the `pathPrefix` option. In this example, the resulting path is `/api/v1/MyPath`.
+
+```typescript
+const c = initContract();
+export const contract = c.router({
+    getPost: {
+      path: 'MyPath',
```
oliverbutler, line 238: Make sure `path` is lowercase and a valid path to avoid confusing folks e.g `/post/:id`

what you've given here will actually be `/api/v1MyPath` at runtime

apps/docs/docs/core/core.md
```diff
@@ -207,3 +224,40 @@ export const contract = c.router({
...
+      //... Your Contract
+    },
+  }, {
+    pathPrefix: '/api/v1',
+  });
+```
+
+You can also use this feature in nested contracts, as shown below. In this case, the resulting path is `/v1/posts/MyPath`, with the `pathPrefix` of the nested contract following the `pathPrefix` of the parent contract.
+
```
oliverbutler, line 247: awesome example but same as above :D

apps/docs/docs/core/core.md
Jwcode-uk: ooooo Whooooops

Jwcode-uk:
@oliverbutler, Thanks for pointing out those mistakes, all fixed now :)

oliverbutler (approved):
Looks awesome! LGTM

Issue #70, opened 2022-10-20
@Api decorator not working properly with nest versioning 


PedroFonseca17:
Hey guys! 
I'm having some issues when creating an endpoint on nestjs + ts-rest that receives a param and has versioning configured by nest have you guys ever faced these issues?

Example: 

Contract route
```ts
 getIndustryTag: {
    method: 'GET',
    path: `/industry-tags/:uid`,
    responses: {
      200: ResponseIndustryTagModel.nullable(),
    },
    summary: 'Get an industry tag',
  },
```
  
  
  Controller
  ```ts
  @Controller({
    version: '1',
  })
  export class IndustryTagsController {
    constructor(private readonly industryTagsService: IndustryTagsService) {}
  
    @Api(server.route.getIndustryTag)
    async findOne(
      @ApiDecorator() { params: { uid } }: RouteShape['getIndustryTag']
    ) {  
          ....
      }
```

The end route for this would be `http://localhost:3000/v1/industry-tags/uid-123` but since the contract route doesn't know about the versioning that was applied, I receive the string `industry-tags` has the param instead of the `uid-123`

Even if I apply the v1 to the contract path it will become `v1/v1/industry-tags` so it wouldn't help either

The only solution I can see working is managing the versioning inside the contract only and letting go of the nestjs versioning system, but I wanted to know if there are other options for this and if this is a known issue

Thanks!

**Comment by oliverbutler** [comment]
This is a super interesting point, we can definitely work on this, the root problem here is that ts-rest is trying to instantiate the path for you, to avoid typos and improve dev x.

Out of interest, would this also work by adding "/v1" to your baseUrl when you call this api?

Or is having the ability to add /v1/.. in the contract important?

**Comment by PedroFonseca17** [comment]
I believe that adding the version to the contract (whole contract and endpoint specific) would be the best approach.
This way, if you are trying to achieve a full type-safety project with API and frontend both of them will know which version of the endpoint they are using just by accessing the contracts information.


So maybe something like this would work:

```
contract.router({
  version: '2':
  getIndustryTag: {
      method: 'GET',
      path: `/industry-tags/:uid`,
      responses: {
        200: ResponseIndustryTagModel.nullable(),
      },
      summary: 'Get an industry tag',
      version: '1'
  },
)}
```

What do you think?

oliverbutler:
I've been playing with this locally for a while, and seeing how it's played out in reality.

[...]

PR #268, opened 2023-06-04
Docs: updated warning as of pathPrefix implementation

Jwcode-uk:
In Nest, global prefix, versioning, or controller prefixes provide options for adding prefixes to paths in your APIs which we do not support.

Previously, there was a warning stating:

'Currently any existing Nest global prefix, versioning, or controller prefixes will be ignored, please see https://github.com/ts-rest/ts-rest/issues/70 for more details.

Currently, the path in your contract must be the **full path**. Your client may specify a different base URL with a prefix, but this cannot be done at the contract level currently.'

However, we have now introduced the ability to add path prefixes. This serves as a workaround for this missing functionality.

I have updated the documentation to reflect this change:

'Currently any existing Nest global prefix, versioning, or controller prefixes will be ignored, please see https://github.com/ts-rest/ts-rest/issues/70 for more details.

We have added the ability to [prefix paths](/docs/core/#path-prefix), allowing more flexibility in defining your API endpoints. This can be used as a workaround for this functionality.'

omg this PR is longer then the change 😂

### DIFF
--- a/apps/docs/docs/nest.mdx
+++ b/apps/docs/docs/nest.mdx
@@ -138,8 +138,5 @@ Currently any existing Nest global prefix, versioning, or controller prefixes wi
 
 If this feature is highly requested, we can investigate a solution.
 
-Currently the path in your contract must be the **full path**, your client may specify a different base url, with a prefix, but this cannot be done at the contract level currently.
-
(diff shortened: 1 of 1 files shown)


Gabrola (approved):
Nice one! 👍

Cheers!

PR #269, opened 2023-06-04
Add new `TsRestHandler` Nest API Proposal

oliverbutler:
- Way less boilerplate! (on PostController in the Nest example anyway)
  - Greatly reduce surface area for human error 
  - Example in the files changed https://github.com/ts-rest/ts-rest/pull/269/files
- Autocomplete in responses (finally no need for `as const`) when you're typing them
- Much more readable errors thanks to not implementing a class to derive types
- It's opt-in so you don't have to convert to this new way


oliverbutler:
One issue in the RC is that a contract with "any" as the body gets messed up

Screenshot 2023-06-06 at 16 24 06

apps/docs/docs/nest/nest.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+However, sometimes it's much cleaner and simpler to throw an error rather than passing a response back to your handler - the problem is that you normally **loose all type safety** when you do this.
+
+With `TsRestException` you can throw a Nest exception and it will be handled by the `tsRestHandler` function and returned as a fully typed response.
+
+```typescript
+throw new TsRestException(contract.test, {
+  status: 400,
+  body: { code: 'UserAlreadySignedUp', message: 'User has already signed up' },
+});
```
Gabrola, line 146: I love this! ❤

apps/docs/src/components/nestTransform.ts
```diff
@@ -0,0 +1,395 @@
+import * as ts from 'typescript';
```
Gabrola, line 1: This is insane 🤯

apps/example-nest/src/app/post.controller.spec.ts
```diff
@@ -76,16 +76,21 @@ describe('PostController', () => {
...
-            {
-              code: 'invalid_type',
-              expected: 'string',
-              message: 'Required',
-              path: ['take'],
-              received: 'undefined',
-            },
-          ],
-          name: 'ZodError',
+          queryResult: {
```
Gabrola, line 79: Do you think it would be better if we kept the `post.controller.ts` and `post.controller.spec.ts` files as-is to make sure we don't have any regressions and create a separate controller and respective test file for the new handlers?

libs/non-strict-mode-test/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "module": "commonjs",
+    "forceConsistentCasingInFileNames": true,
+    "strict": false,
```
Gabrola, line 6: Nice idea to create a separate library without strict mode to test this

libs/ts-rest/nest/src/lib/constants.ts
```diff
@@ -1,3 +1,4 @@
 export const TsRestAppRouteMetadataKey = Symbol('ts-rest-app-route');
+export const TsRestAppRouterMetadataKey = Symbol('ts-rest-app-router');
```
Gabrola, line 2: I think rather than introducing a new metadata key, we can re-use `TsRestAppRouteMetadataKey` and we can easily determine whether it's a multi handler or single handler by checking the value using `isAppRoute`

libs/ts-rest/nest/src/lib/ts-rest-nest-handler.ts
```diff
@@ -0,0 +1,365 @@
...
+      };
+    }
+
+    /**
+     * Therefore we must be on a multi-handler so let's get the route in the router
+     */
+    const appRouter = this.reflector.get<AppRouter | undefined>(
+      TsRestAppRouterMetadataKey,
+      ctx.getHandler()
+    );
```
Gabrola, line 215: I think rather than introducing two separate metadata keys, we can re-use `TsRestAppRouteMetadataKey` and pass `appRouterOrRoute` to the same metadata key in both cases. We can easily determine whether it's a multi handler or single handler by checking the value using `isAppRoute`

libs/ts-rest/nest/src/lib/ts-rest-nest-handler.ts
```diff
@@ -0,0 +1,365 @@
...
+      return null;
+    }) as [string, AppRoute] | undefined;
+
+    if (!foundAppRoute) {
+      throw new NotFoundException("Couldn't find route handler for this path");
+    }
+
+    return {
+      appRoute: foundAppRoute[1],
+      routeAlias: foundAppRoute[0],
```
Gabrola, line 240: This name confused me for a second, maybe `routeKey` would make more sense?

libs/ts-rest/nest/src/lib/ts-rest-nest-handler.ts
```diff
@@ -0,0 +1,374 @@
...
+      paramsResult: pathParams,
+      headersResult: headers,
+      queryResult: query,
+      bodyResult: body,
+    });
+  }
+}
+
+export class ResponseValidationError extends InternalServerErrorException {
+  constructor(appRoute: AppRoute, error: z.ZodError) {
```
Gabrola, line 60: This could help if someone wants to custom handle these using an exception filter

```suggestion
  constructor(public appRoute: AppRoute, public error: z.ZodError) {
```

PR #273, opened 2023-06-12
feat: @ts-rest/serverless package

Gabrola:


.github/workflows/build-test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 16
```
oliverbutler, line 32: Nice catch

apps/example-cloudflare-worker/project.json
```diff
@@ -0,0 +1,36 @@
+{
```
michaelangrivera, line 1: can we add this example in a separate PR?

apps/example-next/app/layout.tsx
```diff
@@ -0,0 +1,22 @@
+import '../pages/styles.css';
+
+import React from 'react';
+import { ReactQueryProvider } from './react-query-utils/react-query-provider';
```
michaelangrivera, line 4: Same here. Can we add this example in a separate PR?

libs/ts-rest/serverless/src/aws.ts
```diff
@@ -0,0 +1 @@
+export * from './lib/handlers/ts-rest-lambda';
```
michaelangrivera, line 1: I'm confused about how consumers will implement this? Will it be a plugin? Will it be also implemented at the handler level (allow users to import utility types/functions for their handlers)? 

Is it possible to generalize this a little further such that something like SST could also use this?

.github/workflows/build-test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
...
+          node-version: 18
```
michaelangrivera, line 32: I thought node 18 wasn't available in GHA yet?

arpadgabor: it actually is available

awilderink:
👀

JoshBarr:
Thanks for making this! We're actually using it successfully in prod. Living life on the edge 😅 

With that in mind, I'd be happy to help move this PR forward if I can be of any assistance?

.github/workflows/build-test.yml
teriu: Definitely available, we've been using it for a while.

libs/ts-rest/serverless/src/aws.ts
Gabrola: It just creates a handler function that can be exported, you can see see some examples in ts-rest-lambda.spec.ts. You can already use this with SST 😉

```typescript
import { createLambdaHandler } from '@ts-rest/serverless/aws';

export const handler = createLambdaHandler(contract, { ... });
```

This creates a handler function that accepts a lambda event and context as arguments `function (event, context)`, which is standard for all AWS lambdas. This is as generic as it gets for AWS Lambdas. If there is a layer on top like SST that defines the handlers in a different format, you can just call that handler function manually and make sure you pass the lambda event and context from wherever it's defined for that specific "framework". So something like this:

```typescript
import { createLambdaHandler } from '@ts-rest/serverless/aws';

const tsRestHandler = createLambdaHandler(contract, { ... });

export const customHandler = async (nonStandardOptions) => {
  return tsRestHandler(nonStandardOptions.event, nonStandardOptions.context);
};
```

libs/ts-rest/serverless/src/lib/mappers/aws/api-gateway.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+export async function responseToResult(
+  event: ApiGatewayEvent,
+  response: TsRestResponse
+): Promise<ApiGatewayResponse> {
+  const headers = {} as Record<string, string>;
+  const multiValueHeaders = {} as Record<string, string[]>;
+
+  response.headers.forEach((value, key) => {
```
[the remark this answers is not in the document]
Gabrola: This is a Web API Headers object not an array so it only has this https://developer.mozilla.org/en-US/docs/Web/API/Headers/forEach

Gabrola:
Hey @JoshBarr! I'll be wrapping this up in the next days. This is awesome that you are already using this successfully in prod (I hope until now 😆). If you've come across anything peculiar or you think there is some improvement or change that needs to be made, please let me know.

libs/ts-rest/serverless/src/aws.ts
JoshBarr: SST just uses the standard `Handler` type from `aws-lambda`. We're using `createLambdaHandler` directly in our applications. Works well.

JoshBarr:
@Gabrola We had to make some changes to the module output format to get it building inside our typescript/SST apps. 

Here's the patch file:
https://gist.github.com/JoshBarr/2b64ed796cea542743de40e165abc10e

JoshBarr:
Also, I am happy to assist with getting this work completed - what can I do? 😀

kyleamazza:
Also interested in helping get this implemented! Is there anything that would help to get this in a good state for approval (i.e. moving examples to separate PRs)?

libs/ts-rest/serverless/src/aws.ts
abubakar-vaival: can you give more example of SST

abubakar-vaival: kindly help

PR #291, opened 2023-06-28
feat: @ts-rest/vue-query

### LINKED ISSUE #71: Feature request - query adapter support (vue)
add use query adapters support for example vue - https://tanstack.com/query/v4/docs/adapters/vue-query

fruchtzwerg:
resolves #71 

Mostly copy-paste from @ts-rest/solid-query lib. I substituted @tanstack/solid-query with @tanstack/vue-query where necessary.

See `pnpm nx serve example-microservice-web-app-vue` for example usage.


apps/docs/docs/vue-query.mdx
```diff
@@ -0,0 +1,176 @@
...
+const queryResult = client.posts.get.useQuery(
+  ['posts'], // <- queryKey
+  { params: { id: '1' } }, // <- Query params, Params, Body etc (all typed)
+  { staleTime: 1000 } // <- react-query options (optional)
+);
+```
+
+:::tip
+
+The design philosophy of `@ts-rest/vue-query` is to make it as easy as possible to use `react-query` with `@ts-rest`. This means that the `useQuery` and `useMutation` hooks are as close to the original `react-query` hooks as possible, as such we don't abstract away from the query keys or the options. Only the query function itself!
```
michaelangrivera, line 67: should this say `react-query`? Shouldn't it be `vue-query`?

fruchtzwerg: I was wondering myself. Technically it should probably be TanStack Query in the docs for both `@ts-rest/react-query` and `@ts-rest/vue-query`. I wanted to keep it consistent without changing the existing docs.

PR #299, opened 2023-07-06
feat: @ts-rest/vue-query

### LINKED ISSUE #71: Feature request - query adapter support (vue)

fruchtzwerg:
resolves #71 

Let's try this again.

I've implemented the example app against `apiBlog` this time, to make sure parameterised queries and infinite query work as expected.
Also updated the docs to reference `vue-query` instead of `react-query`.

Mostly copy-paste from `@ts-rest/react-query` lib. I substituted `@tanstack/react-query` with `@tanstack/vue-query` where necessary.

See `pnpm nx serve example-microservice-web-app-vue` for example usage.


Gabrola:
@fruchtzwerg this looks awesome! Thank you for your contribution ❤
We'll review this over the coming days hopefully. In the meantime, would it be possible to add some tests? Have a look at how we're running tests for react here https://github.com/ts-rest/ts-rest/blob/main/apps/example-next/tests/react-query.spec.tsx

I'm not familiar with vue that much, but I believe there are vue testing utils that should help with this in the `@testing-library/vue` library,

fruchtzwerg:
@Gabrola sure thing. I'm on vacation for the next two weeks so I'll look into it when boredom strikes.
I'll look out for abstraction opportunities. Maybe some tests could be reused between the tanstack query libs.

fruchtzwerg:
Added the relevant tests from the example-next.
I kept the ones for missing features as skipped tests to demonstrate what is not yet supported by vue-query.

To keep it simple, I implemented a stub of `renderHook` from `@testing-library/react-hooks` for Vue, thus allowing to copy-paste tests between Vue and React.

The only difference between the two is that Vue's reactivity system requires values to be wrapped in a `Ref` object. This means that they have to be `unref`-ed prior to comparison with the expected value. It is possible to do this globally in a Jest setup script. That would allow both `@ts-rest/react` and `@ts-rest/vue` to run against the same tests.
I have not included this here because I'm unsure how well this would scale with solid and other integrations you might want to support in the future.
If you're interested I can iron out the kinks and provide the commit.

oliverbutler (commented):
This looks great thanks! Threw me at first using example-nest's API for the `example-microservice` example set 

Maybe worth us standardising our example apps as we grow - but this looks great thanks!

Let's get it merged asap, I'm sure theres other nitpicks we can go over once the conflicts are resolved and we're in sync with master

fruchtzwerg:
Thanks for the review @oliverbutler.

I resolved the conflicts due to dependencies.
Tanstack Query will remove overloads of `useQuery()` in [v5](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5). I updated the implementation to maintain compatibility when it releases.

fruchtzwerg:
@michaelangelo-io I'm unfamiliar with changesets so I hope this works. I also squashed my commits.

michaelangelo-io (approved):

fruchtzwerg:
Docs build seems to be broken. I merged @oliverbutler's latest changes on main but it's still complaining about `Readme.md`. Something about the lonely `</br>` missing a corresponding open tag and the `style="padding-top: 25px"` not being valid JSX. Seems fishy to me; this should still be valid HTML ¯\_(ツ)_/¯

Since this is blocking this repo and I'm not too familiar with Docusaurus, I'll leave it to you to fix this on main.

oliverbutler:
@fruchtzwerg You hit the nail on the head - it was docusaurus stalling (not exiting gracefully upon failure) which is super annoying. 

I hope we get free actions billing...

oliverbutler:
@fruchtzwerg if you update from master it'll pass now 🤞 sorry about that 🤦 This is what I get for making _one change to master_ to README

fruchtzwerg:
@oliverbutler No worries, this is ~pod racing~ software development after all.
I think we should be good now. I merged the latest version and ran `pnpm build-test-all` locally and its all 🟢 now.

As you mentioned earlier there are probably some minor bugs and it doesn't implement the full feature-set of tanstack query but it should cover about 80-90% of the most common usecases.

oliverbutler:
@all-contributors please add @fruchtzwerg for code, docs, and test

PR #314, opened 2023-07-23
feat: expose internal types for @ts-rest/express

### LINKED ISSUE #312: expose internal types in `./lib/types` for `@ts-rest/express`
https://github.com/ts-rest/ts-rest/blob/c90d2ce9ff415bc72284d1ba937919b279e74448/libs/ts-rest/core/src/index.ts#L1-L11

https://github.com/ts-rest/ts-rest/blob/c90d2ce9ff415bc72284d1ba937919b279e74448/libs/ts-rest/express/src/index.ts#L1-L3

unlike `@ts-rest/core`, `@ts-rest/express` does not expose its router types in `.lib/types`.

https://github.com/ts-rest/ts-rest/blob/c90d2ce9ff415bc72284d1ba937919b279e74448/libs/ts-rest/express/src/lib/types.ts#L46-L53

It would be very convenient to extract express router functions from one giant router if it was possible to use utility types like `AppRouteOptions` directly.

scarf005:
# Summary

- resolves #312 

exposed useful internal types like `AppRouteOptions`.

## Example usage

```ts
import { AppRouteOptions } from '@ts-rest/express'

export type HandlerFor<T extends AppRoute> = AppRouteOptions<T>['handler']

// ...

type PatchDeps = Pick<ReviewsService, 'patchReviews'>
type MkPatch = (deps: PatchDeps) => HandlerFor<typeof contract.reviews.patch>
export const mkPatchReviews: MkPatch =
  ({ patchReviews }) =>
  async ({ params: { reviewsId }, req: { user } }) => {
    const { id: userId } = getUser.parse(user)
    const result = await patchReviews({ reviewsId, userId })

    return match(result)
      .with(P.instanceOf(ReviewNotFoundError), () => reviewNotFound)
      .otherwise(() => ({ status: 200, body: '리뷰 공개 여부가 업데이트되었습니다.' } as const))
  }

// ...
const postHandler = mkPostReviews({ createReviews })

export const reviews = s.router(contract.reviews, {
  post: {
    middleware: [authValidate(roleSet.all)],
    handler: postHandler,
  }
})
```

### DIFF
--- a/libs/ts-rest/express/src/index.ts
+++ b/libs/ts-rest/express/src/index.ts
@@ -1,3 +1,8 @@
 export * from './lib/ts-rest-express';
-export { TsRestRequest, TsRestRequestHandler } from './lib/types';
+export type {
+  AppRouteImplementation,
+  AppRouteOptions,
+  TsRestRequest,
+  TsRestRequestHandler,
+} from './lib/types';
 export { RequestValidationError } from './lib/request-validation-error';
--- a/.changeset/big-colts-carry.md
+++ b/.changeset/big-colts-carry.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/express': patch
+---
+
+Exposed `AppRouteOptions` and `AppRouteImplementation` types. They can be used to split router handlers.
--- a/.changeset/big-colts-carry.md
+++ b/.changeset/big-colts-carry.md
@@ -2,4 +2,4 @@
 '@ts-rest/express': patch
 ---
 
-Exposed `AppRouteOptions` and `AppRouteImplementation` types. They can be used to split router handlers.
+Exposed `AppRouteOptions` and `AppRouteImplementation` types. They allow for greater developer flexibility and can be used to split router handlers.


libs/ts-rest/express/src/index.ts
```diff
@@ -1,3 +1,3 @@
 export * from './lib/ts-rest-express';
-export { TsRestRequest, TsRestRequestHandler } from './lib/types';
+export * from './lib/types';
```
michaelangelo-io, line 2: Hmm, can we instead expose particular types that make sense? 
By exposing all, this leaves room for future headache should we want do any internal refactoring, since that would potentially break downstream consuming clients.

scarf005: I see. how about only exposing `AppRouteImplementation` and `AppRouteOptions` types only? they would be useful for extracting individual handler functions.

michaelangelo-io (approved):

libs/ts-rest/express/src/index.ts
michaelangelo-io: SGTM!

PR #315, opened 2023-07-23
WIP (Next 13 app Dir fetch support!)

michaelangelo-io:


libs/ts-rest/core/src/lib/infer-types.ts
```diff
@@ -24,6 +24,12 @@ import {
 import { ApiFetcher, ClientArgs } from './client';
 import { ParamsFromUrl } from './paths';
 
+export type Frameworks = 'nextjs' | 'none';
+
+export type NextClientArgs = {
+  next?: { revalidate?: number | false, tags?: string[]} | undefined
+}
```
michaelangelo-io, line 31: I couldn't think of a smart way to create this type from the `next-env.d.ts` file without creating `next` as an implicit dependency. Lmk if y'all have ideas 😬 

We may have to maintain this unfortunately, but luckily it's not too much.

rklubenspies: @michaelangelo-io It may also be worth noting that Next's App Router caches everything by default and this is not desired functionality in apps that deal with user-specific data being returned at common routes (like a dashboard route). So to avoid information leakage, developers will often set the `cache` option on the `fetch` request to `no-store`. The cache option is native to the fetch function but I believe it is not implemented in the default `tsRestFetchApi` exposed by `@ts-rest/core`. To support App Router's full implementation of fetch, it might be necessary need to add a `cache` key to the `NextClientArgs` type above the `next?` option and also pass both `cache` and `next` arguments through to the underlying `fetch` function in `tsRestFetchApi`.

michaelangelo-io: Great idea, I'll add the cache as an option as well. We should definitely be exposing that as that is a standard web API now.

michaelangelo-io: Done !

libs/ts-rest/core/src/lib/client.ts
```diff
@@ -152,6 +167,12 @@ export const fetchApi = ({
   extraInputArgs: Record<string, unknown>;
   headers: Record<string, string | undefined>;
   signal?: AbortSignal;
+  // ---- Framework specific ----
+  /**
+   * only to be used by @ts-rest/next
+   * You can obtain a Nextjs Client by calling `initNextClient`
+   */
+  next?: NextClientArgs['next'] | undefined;
```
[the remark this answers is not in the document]
oliverbutler: Chatted in Discord 👍 :))

oliverbutler (approved):
LGTM! Spotted an issue in example-microservice/web-app (type error) but this seems to be an existing problem not caused by you

oliverbutler (approved):
LGTM! Spotted an issue in example-microservice/web-app (type error) but this seems to be an existing problem not caused by you

PR #326, opened 2023-07-31
fix: return errors from fastify route handlers

suphon-t:
`requestValidationErrorHandler` currently only handles `RequestValidationError` while silently consuming all other errors. This means that if there is any unhandled error in a route handler, the request never completes.

I added a test case for demonstration, and fixed the issue by re-throwing errors other than `RequestValidationError`.

### DIFF
--- a/libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
+++ b/libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
@@ -176,6 +176,8 @@ const requestValidationErrorHandler = (
       } else {
         return handler(err, request, reply);
       }
+    } else {
+      throw err;
     }
   };
 };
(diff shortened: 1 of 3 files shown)


libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
```diff
@@ -176,6 +176,8 @@ const requestValidationErrorHandler = (
       } else {
         return handler(err, request, reply);
       }
+    } else {
+      throw err;
```
gustawdaniel, line 180: I receiving here

```
http: LogLevel.ERROR: ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response')) while doing a GET request to URL: http://localhost:4000/
```

I prefer to get 500 with error message or even better be able to override it by my own handler.

gustawdaniel (changes requested):

gustawdaniel (approved):

michaelangelo-io (approved):
LGTM!

michaelangelo-io:
Can you add a changeset bump to the fastify package @suphon-t?

PR #338, opened 2023-08-08
fix: tsRestFetchApi content type parsing

### LINKED ISSUE #335: Response does not parse when response content-type header is `application/problem+json`
Trying out the library, and noticed this.

APIs built on `.NET` sometimes (tend to) return a `content-type` header value of `application/problem+json; charset=utf-8`.
This header value can be seen on a `400` level error.

This presents a problem as the `tsRestFetchApi` function only uses the `.json()` parse method when the content-type header strictly matches `application/json` as seen [here](https://github.com/ts-rest/ts-rest/blob/main/libs/ts-rest/core/src/lib/client.ts#L94-L100).

Not sure if this is something to be covered within the scope of this project, but it's something I ran into, and I'll leave it up to you folk to decide on.
For now, I'll be rolling my own fetch adapter allowing for this variation of the header's value whilst testing out the library.

michaelangelo-io:
fixes #335

### DIFF
--- a/.changeset/four-days-wonder.md
+++ b/.changeset/four-days-wonder.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/core': minor
+---
+
+tsRestFetchApi should be more flexible when determining application/json content type header
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -91,7 +91,7 @@ export const tsRestFetchApi: ApiFetcher = async ({
   });
   const contentType = result.headers.get('content-type');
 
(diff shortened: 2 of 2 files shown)


oliverbutler (approved):
I would've loved to see a test for this - but to be frank this is a pretty trivial change so I'm not worried

michaelangelo-io:
> I would've loved to see a test for this - but to be frank this is a pretty trivial change so I'm not worried

yup, I think it should be covered in https://github.com/ts-rest/ts-rest/blob/d122bb7dffc17c1a4bb5bde184ab0850c9f31ea4/libs/ts-rest/core/src/lib/client.spec.ts

Issue #349, opened 2023-08-12
Type error when combining server contracts with fastify


ryanstaniforth:
After failing to figure out how to combine routers, I found issue #262. Turns out I had it right but there seems to be an issue with `@ts-rest/fastify`.

This code presents a type error at `auth: authServer`, but compiles fine if you swap `@ts-rest/fastify` for `@ts-rest/express`.

```typescript
import { initContract } from '@ts-rest/core';
import { initServer } from '@ts-rest/fastify';
import { z } from 'zod';

const c1 = initContract();

export const authContract = c1.router({
  login: {
    method: 'POST',
    path: `/auth/login`,
    body: z.object({
      email: z.string(),
      password: z.string(),
    }),
    responses: {
      200: z.object({
        token: z.string(),
      }),
    },
  },
});

const c = initContract();

export const contract = c.router({
  auth: authContract,
});

const s1 = initServer();

export const authServer = s1.router(contract.auth, {
  login: async () => {
    return {
      status: 200,
      body: {
        token: 'secret token',
      },
    };
  },
});

const s = initServer();

export const router = s.router(contract, {
  auth: authServer,
});
```

oliverbutler:
Hey @ryanstaniforth thanks for reporting - this is bizarre upon first glance.

I'll take a look at the differences between the two libs, it's very possible I missed testing combined contracts with the fastify lib...

oliverbutler:
As I'm investigating this and writing some more tests around the lib I want to suggest a super temporary fix 

I'd (for now) register each router individually 

```ts
const contractA = c.router({
      a: {
        method: 'GET',
        path: '/a',
        responses: {
          200: c.otherResponse({
            contentType: 'text/plain',
            body: c.type<string>(),
          }),
        },
      },
    });

    const contractB = c.router({
      b: {
        method: 'GET',
        path: '/b',
        responses: {
          200: c.otherResponse({
            contentType: 'text/plain',
            body: c.type<string>(),
          }),
        },
      },
    });

    const routerA = s.router(contractA, {
      a: async () => {
        return {
          status: 200,
          body: 'a',
        };
      },
    });

    const routerB = s.router(contractB, {
      b: async () => {
        return {
          status: 200,
          body: 'b',
        };
      },
    });

    const app = fastify({ logger: false });

    app.register(s.plugin(routerA)).register(s.plugin(routerB));

    await app.ready();
```

hopefully a workaround whilst i investigate!

[...]

PR #351, opened 2023-08-13
fix: unable to nest fastify route implementations

oliverbutler:
issue explained in #349

### DIFF
--- a/.changeset/mighty-spoons-float.md
+++ b/.changeset/mighty-spoons-float.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/fastify': patch
+---
+
+fix: Type error when combining server contracts with fastify
--- a/libs/ts-rest/fastify/src/lib/ts-rest-fastify.spec.ts
+++ b/libs/ts-rest/fastify/src/lib/ts-rest-fastify.spec.ts
@@ -110,7 +110,7 @@ describe('ts-rest-fastify', () => {
   it('should instantiate fastify routes using plugin instance', async () => {
     const app = fastify({ logger: false });
 
-    app.register(s.plugin(contract, router));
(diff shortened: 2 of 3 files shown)


oliverbutler:
@Gabrola @michaelangelo-io (@baryla adding you here as you implemented this nested structure initially)

The issue here is stumping me, we're returning 

```ts
type InitialisedRouter<TContract extends AppRouter> = {
  contract: TContract;
  routes: RecursiveRouterObj<TContract>;
};
```

from a `.router` in fastify.

The problem here is when you try to embed a router in a router you get an error as 

```ts
type InitialisedRouter<TContract extends AppRouter> = {
  contract: TContract;
  routes: RecursiveRouterObj<TContract>;
};
```

is NOT assignable to the expected 

```ts
RecursiveRouterObj<TContract>;
```

for the implementation in the route.

This leads to the fix for 

```ts
const routerA = s.router(contractA, {
      a: async () => {
        return {
          status: 200,
          body: {
            a: 'return',
          },
        };
      },
    });

    const routerB = s.router(contractB, {
      b: async () => {
        return {
          status: 200,
          body: {
            b: 'return',
          },
        };
      },
    });

    const combinedRouter = s.router(combinedContract, {
      apiA: routerA.routes, // <- The pretty dirty fix
      apiB: routerB.routes, // <- The pretty dirty fix
    });
```

oliverbutler:
After playing with this more and making a non-breaking version of the fix I actually like the `ImplementedRouter` design.

It could let us add a second parameter to `.router` for options like we do in the Nest library - a feature i believe to be missing from the fastify routes.

michaelangelo-io (approved):
LGTM

PR #372, opened 2023-08-29
feat: add better typing for TsRestRequest

tefkah:
This PR adds better type inference for the `req` object of route implementations when using `@ts-rest/express`.

Consider this minimal example, modified from here

https://github.com/tefkah/ts-rest/blob/2073f362cbb1eb35a8bd06e65f36494dc2d8ed8c/libs/ts-rest/express/src/lib/ts-rest-express.spec.ts#L99-L104

Before this PR, after adding `req` to the destructuring argument like so 
```ts
      postIndex: async ({ body: { echoHtml }, req }) => {
```

it would have type `any`

```ts
      postIndex: async ({ body: { echoHtml }, req }) => {
			const body = req.body
				  //  ^? const body: any
```

After this PR, it will have type `{ echoHtml: string }`, i.e.
```ts
      postIndex: async ({ body: { echoHtml }, req }) => {
			const body = req.body
				  //  ^? const body: { echoHtml: string }
```						

Also `req.headers`, `req.query`, and `req.params` will be typed.

I'm not sure what the best practice is to test this, given that there weren't any type tests in `libs/ts-rest/express/src/lib`, so I added one. Feel free to remove it if it, it's more for demonstration purposes.

I also tried to type the `res` object better, but ran into problems there for cases where `strictStatusCodes` is `false` or when there are multiple statuscodes defined. Typescript will try to merge the statuscodes/response jsons like [in this stackoverflow question](https://stackoverflow.com/questions/75239505/using-unions-to-type-express-response-body-inside-generic-argument-or-outside), which leads to a very bad DX. 

Let me know what you think!

### DIFF
--- a/libs/ts-rest/express/src/lib/types.ts
+++ b/libs/ts-rest/express/src/lib/types.ts
@@ -4,9 +4,15 @@ import {
   AppRouteQuery,
   AppRouter,
   ServerInferRequest,
+  ServerInferResponseBody,
   ServerInferResponses,
 } from '@ts-rest/core';
-import { Express, NextFunction, Response } from 'express-serve-static-core';
+import {
+  Express,
+  NextFunction,
+  Response,
+  Request,
+} from 'express-serve-static-core';
 import { RequestValidationError } from './request-validation-error';
 
 type AppRouteQueryImplementation<T extends AppRouteQuery> = (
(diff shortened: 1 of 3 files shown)


tefkah:
I see a changeset is needed, I assume I should not be the one to add it right?

libs/ts-rest/express/src/lib/types.ts
```diff
@@ -32,10 +38,21 @@ export type AppRouteImplementation<T extends AppRoute> =
     ? AppRouteQueryImplementation<T>
     : never;
 
-export type TsRestRequest<T extends AppRouter | AppRoute> =
-  Express['request'] & {
-    tsRestRoute: FlattenAppRouter<T>;
-  };
+export type TsRestRequest<
+  T extends AppRouter | AppRoute,
+  F extends FlattenAppRouter<T> = FlattenAppRouter<T>,
```
tefkah, line 43: These could also be defined in the body of this definition with some type helpers or something if you do not prefer this.

michaelangeloio:
This LGTM!

Can you add a changeset with `npx changeset`? For a minor version of express.

libs/ts-rest/express/src/lib/types.ts
michaelangeloio: I think this is fine, no need to over abstract for now.

michaelangeloio:
> I see a changeset is needed, I assume I should not be the one to add it right?

Yes please add one via `npx changeset` to a minor version of express. I'll modify it if needed before merging to master!

michaelangeloio (approved):

Issue #306, opened 2023-07-19
Feature request: client-side schema validation option


GabeAtWork:
# Summary

This issue is a request for a new feature: allow the client to automatically run the contact's schemas on the body upon receiving a response via a new option in `initClient`.

Rationale:
- If the server/client get out of sync with a breaking change, this would help it be caught when the query runs, as early as possible
- The schemas can operate some transformation (e.g. decoding a string date), alleviating the need for the client code to write and apply their own schema

# Proposed spec

Add a new option `validateBodyWithContract` to `ClientArgs`:
```ts
export interface ClientArgs {
  baseUrl: string;
  baseHeaders: Record<string, string>;
  api?: ApiFetcher;
  credentials?: RequestCredentials;
  jsonQuery?: boolean;
  // New option here
  validateBodyWithContract?: boolean;
}
```

If the route's definition includes a schema for the response code (i.e. in my understanding, if the contract is `z.ZodTypeAny | ContractPlainType<unknown>`, which we may potentially want to simplify to the contract having a `parse` method), it gets used to parse the body in `getRouteQuery`.

# Usage

One could enable this simply by doing:

```ts
const client = initClient(contract, {
  validateBodyWithContract: true
})
```

# Alternatives

In order to solve the same problem, we could also:
* Add an option on the route definition in the contract (upside: more granular, downside: the contract now decides whether the validation should run or not)
* Add a new option `initClient` to enable/disable this behaviour on a per-route basis (upside: more granular, downside: more complex, may be kept as a future improvement)

# Misc

Here's approximately what the implementation could look like:

```ts
if (knownResponseStatuses.includes(response.status.toString())) {
      // TODO make it work with throwOnUnknownStatus disabled
      if(clientArgs.validateWithSchemas) {
        const routeResponseSchema = route.responses[response.status];
        if(!routeResponseSchema || !("parse" in routeResponseSchema)) {
          console.warn('Route schema does not have a parse method, cannot parse', {
            status: response.status
          });
          return response;
        }

        const parsedBody = routeResponseSchema.parse(response.body);

        return {...response, body: parsedBody};
      }

      return response;
    }
```

SeanCassiere:
Got my 👀 on this one.

Was porting over an app yesterday to ts-rest when I noticed that my zod validation schema wasn't really being called since I had a transform on a field which was never applied.

Basically led to Typescript thinking a property in my response was of a certain type, which it wasn't.

stychu:
[...]

PR #373, opened 2023-08-30
feat: add response validation to the client

SeanCassiere:
Addresses #306.

I'm not sure how you'd want testing to be approached for this, as such it's being left open-ended.
The current test suite run using `pnpm build-test-all` executes all the tests successfully.

Overview of changes: 
* Added an optional `boolean` field named `validateResponseOnClient` to the `AppRouteCommon` and `RouterOptions` interfaces to allow for conditional route or router-based response validation.
* Changed the type of `z.ZodTypeAny` to `z.ZodSchema` on `ContactAnyType`.
* Added runtime validation for `application/json` responses to the `tsRestFetchApi`.

### DIFF
--- a/apps/docs/docs/core/core.md
+++ b/apps/docs/docs/core/core.md
@@ -228,6 +228,23 @@ export const contract = c.router({
 });
 ```
 
+:::caution
+When using `zod` as the schema, should the validation fail, the error will be thrown as a `ZodError`.
+
+You can catch this error and handle it however you like.
+
+```typescript
+try {
+  const posts = await client.getPosts();
+} catch (error) {
+  if (error instanceof ZodError) {
+    // handle error
+  }
+}
+```
+
+:::
+
 ## Metadata
 
 You can attach metadata with any type to your contract routes that can be accessed anywhere throughout ts-rest where
--- a/apps/docs/docs/core/core.md
+++ b/apps/docs/docs/core/core.md
@@ -90,13 +90,16 @@ This will force the client to always pass
 
 ```typescript
 const c = initContract();
-export const contract = c.router({
-  // ...endpoints
-}, {
-  baseHeaders: z.object({
-    authorization: z.string(),
-  }),
-});
+export const contract = c.router(
+  {
+    // ...endpoints
+  },
+  {
+    baseHeaders: z.object({
+      authorization: z.string(),
+    }),
+  }
+);
 ```
 
 ## Responses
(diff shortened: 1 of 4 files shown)


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -99,16 +100,31 @@ export const tsRestFetchApi: ApiFetcher = async ({
...
+    const statusCode = result.status;
+    const response = route.responses[statusCode];
+
     return {
-      status: result.status,
-      body: await result.json(),
+      status: statusCode,
+      body:
+        response && typeof response !== 'symbol' && 'parse' in response
+          ? response?.parse(jsonData)
```
michaelangeloio, line 126: are we sure we want to do `parse` instead of `safeParse`? I feel like we should allow the client consumers to have more flexibility in their error handling? Or maybe it makes sense to just throw and let the client catch the error?

SeanCassiere: `safeParse` could be used here, but that presents a couple problems.
1. What happens if it doesn't pass validation? What happens to the flow then?
2. How do you convey the details of the error to the client? (For those that either want to decorate their forms or show a toast/banner)

Unless the plan is to offer some kind of `clientOnError` function (which probably wouldn't be feasible since the contract also needs to be compatible for sharing on the server, and the contract API surface would be broadening), you'd then probably have to throw a custom TsRest flavoured Error.
If you do go down the custom Error thrown approach, you'd need to have properties on it for returning the original error data and to distinguish the type of validator being used (i.e: zod, internal, custom, ....).

In its current implementation, you are able to check if the error thrown is an `instanceof ZodError` and deal with it accordingly.

That was my rationale for using the 'parse' fn with zod.

michaelangeloio: I completely agree, just wanted to make sure we had this discourse. 

Can you add some documentation to this feature in this PR or a subsequent PR and mention that it will throw (and you can check for `instanceof ZodError`)?

SeanCassiere:
@michaelangeloio Let me know if I messed up the changeset... Never used it before .

libs/ts-rest/core/src/lib/client.ts
SeanCassiere: Sure thing!

apps/docs/docs/core/core.md
```diff
@@ -165,6 +168,38 @@ export const contract = c.router({
...
+
+```typescript
+const c = initContract();
+export const contract = c.router({
+  getPosts: {
+    ...,
+    validateResponseOnClient: true,
+  }
+});
+```
```
michaelangeloio, line 201: Awesome! Cheers 🍾

michaelangeloio (approved):

apps/docs/docs/core/core.md
SeanCassiere: I seem to have made a teeny tiny oopsie here.

My caution is a sub-heading down. Fixing now.

PR #375, opened 2023-08-30
fix: release action fix by reverting to older @nx/rollup version

michaelangeloio:
For some reason, any version of `@nx/rollup` higher than 16.5.5 seems to break our CI by outputting compiled files as `cjs` instead of `js`. 

This is a temporary workaround to unblock releases until a long term strategy is figured out. 

https://github.com/nrwl/nx/issues/18912

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@nestjs/platform-express": "10.1.2",
     "@nestjs/platform-fastify": "10.1.2",
     "@nestjs/swagger": "7.1.4",
-    "@nx/rollup": "16.7.4",
+    "@nx/rollup": "16.5.5",
     "@prisma/client": "^4.5.0",
     "@react-navigation/native": "^6.0.13",
     "@react-navigation/stack": "^6.3.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -57,8 +57,8 @@ importers:
         specifier: 7.1.4
         version: 7.1.4(@nestjs/common@10.1.2)(@nestjs/core@10.1.2)(reflect-metadata@0.1.13)
       '@nx/rollup':
-        specifier: 16.7.4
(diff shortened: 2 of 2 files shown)


oliverbutler (approved):

llllvvuu:
Hi, the `@ts-rest/express` build is actually broken with <=16.5.5.

The reason is that this PR https://github.com/nrwl/nx/pull/18342 added rolling up of the *.d.ts so that all types are exported.

Without it, turning on `declaration: true` will cause users to attempt to re-export inferred types from `@ts-rest/express/lib/types`, which is *not* exported by `@ts-rest/express`, thus causing a build error:

```
src/routes/getOrder.ts:7:14 - error TS2742: The inferred type of 'getOrder' cannot be named without a reference to '../../node_modules/@ts-rest/express/src/lib/types.js'. This is likely not portable. A type annotation is necessary.

7 export const getOrder = s.router(contract.getOrder, {
               ~~~~~~~~
```

There are many situations in which one would want to turn on `declaration: true`:
- One is using `@ts-rest/express` in another library
- One is using `@ts-rest/express` in an app but wanting to split `src/` and `tests/` to make compilation/IDE on `src/` faster. This requires TypeScript project references, i.e. emit the `src/` `.d.ts` so that `tests/` can use it

Our current workaround is monkey-patching the generated `package.json`:
```diff
"exports": {
  ".": {
    "types": "./src/index.d.ts",
    "import": "./index.mjs",
    "require": "./index.js"
  },
+  "./lib/types": {
+    "types": "./src/lib/types.d.ts"
+  }
},

```

michaelangeloio:
hey @llllvvuu ! Do you think this is an appropriate issue for tracking this fix? https://github.com/ts-rest/ts-rest/pull/375

llllvvuu:
Yeah, that looks like it might be the same issue. I'm not sure why the author closed it though. Maybe they realized they didn't need `"declaration": true`?

PR #387, opened 2023-09-11
fix: content-type application/json behavior for client

### LINKED ISSUE #383: DELETE with react-query client breaks fastify because it sets content-type request header to application/json
The ts-rest client with react-query erroneously sets the request header "content-type: application/json" for `DELETE` requests that appropriately have an empty body.

I have not tested with the fetch client however it may be impacted as well.

This header behaviour breaks Fastify with its default content type parser. Clients receive an error 500 response.

Internally the error is logged as `FST_ERR_CTP_EMPTY_JSON_BODY`: https://fastify.dev/docs/latest/Reference/Errors/#fst_err_ctp_empty_json_body

This behaviour can be replicated with any API client such as Postman or the VSCode ThunderClient extension: 

- if you send a valid DELETE request to your ts-rest powered Fastify API (e.g. to an endpoint with 'DELETE' method, no body, and receiving an id via params) then it will be accepted fine
- if you add "Content-Type" "application/json" to the request headers and send a similar DELETE request with no body you will produce the same error

The ts-rest client is replicating this exact same error condition with its behaviour of setting "content-type" (note: it looks exactly like that in my network tab, specified in lowercase).

In my contract definition with `method: 'DELETE'` I have `body: null`. If this should be set to something else or a particular zod schema (e.g. `z.void()`) I think that should be clear in the docs (as the typing requires that a `body` property exist) and there should be test cases surrounding this.

---

An undesirable workaround is setting the `body` of the `method: 'DELETE'` contract to `z.object({})` in the contract definition and then specifying an empty object (`{}`) as the body on the client-side in React.

I don't think this workaround should be relied on as I think it defeats the purpose of ts-rest to implement standard RESTful API's that behave the way external/non-ts-rest-client users expect.

---

Issue #307 may be related to this in some way. It is not clear from the OP if they are using Fastify or not.

In researching this issue I found the same problem was faced by React-Admin in the past. Check out the following for context:
https://github.com/zachrybaker/ra-data-rest-client/issues/1

Fastify has had a few related issues here and there in the past e.g. https://github.com/fastify/fastify/issues/297 (see links to other mentions). My understanding is that the maintainers have concluded that the behaviour described in this issue is correct and the error response is warranted.

Thanks for your attention on this issue!

michaelangeloio:
fixes #383 

Please review: According to [this thread](https://stackoverflow.com/questions/5661596/do-i-need-a-content-type-header-for-http-get-requests), for specifying `content-type: application/json` should be _optional_ for GET requests. 

Will this be ok so long as we correctly communicate this change?

### DIFF
--- a/.changeset/popular-ants-relax.md
+++ b/.changeset/popular-ants-relax.md
@@ -1,0 +1,6 @@
+---
+'@ts-rest/core': patch
+'@ts-rest/vue-query': patch
+---
+
+fix: content-type application/json will not be automatically included in client request if the route is GET or body is undefined/null
--- a/apps/example-next/tests/react-query.spec.tsx
+++ b/apps/example-next/tests/react-query.spec.tsx
@@ -184,7 +184,6 @@ describe('react-query', () => {
       path: 'https://api.com/health',
       body: undefined,
       headers: {
-        'content-type': 'application/json',
         'x-test': 'test',
       },
       route: router.health,
@@ -225,7 +224,6 @@ describe('react-query', () => {
       path: 'https://api.com/posts/1',
       body: undefined,
       headers: {
-        'content-type': 'application/json',
         'x-test': 'test',
       },
       route: router.posts.getPost,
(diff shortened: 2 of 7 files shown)


Gabrola (approved):
LGTM!

PR #413, opened 2023-10-16
fix: allow blob form uploads to support node 18

tefkah:
This PR changes one line, changing the `instanceof File` check to `instanceof Blob` in form uploads.

This makes form uploads possible with the generated client in `node` 18. Previously it would crash and throw something like

```sh
    ReferenceError: File is not defined

      40 |               rest: Record<string, unknown> | undefined
      41 |             ]
    > 42 |             return value.call(this, {
         |                          ^
      43 |               body: { ...body, communityId },
      44 |               ...rest,
      45 |             })
```

This also improves compatibility with `node`'s `FormData`, which allows for `Blob`s to be added to `FormData`, as well as allowing you to submit forms using `node` 18.

I do see that you have [a test that uses `File` explicitly here](https://github.com/ts-rest/ts-rest/blob/984b9b5aa229eb60562e7805a3f4bc8666ed0a26/libs/ts-rest/core/src/lib/client.spec.ts#L523-L527), I don't know why that works here, maybe some `jest` settings you are using?

### DIFF
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -148,7 +148,7 @@ const createFormData = (body: unknown) => {
   const formData = new FormData();
 
   Object.entries(body as Record<string, unknown>).forEach(([key, value]) => {
-    if (value instanceof File) {
+    if (value instanceof Blob) {
       formData.append(key, value);
     } else {
       formData.append(key, JSON.stringify(value));


Gabrola:
The problem with `Blob` is that it does not have a file name so when it is added to the `FormData`, the file name will default to `blob` and with how ts-rest is currently setup there will be no way to set the filename. `fetch` support is currently experimental hence why not all of the web APIs are still available, in this case `File`.

I'm not 100% if this works, but I'm confident enough that if you globally polyfill `File` using the `@web-std/file` package then your issue will be fixed

tefkah:
Thanks for your response! 

I do get that you cannot set a filename to a Blob, but there are different ways of supplying filenames for e.g. an upload, e.g. as a separate form field.
I do feel like it makes sense to allow Blob uploads, [since this is also supported in the browser](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append). I personally do not think `ts-rest` should "dictate" that only Files are allowed to be used in forms. (of course my actual reason is that it would be easier for me this way haha)

Maybe an API for adding filenames could be added, e.g. you could maybe pass a tuple of a blob and a string which which would be used in `formData.append`. I do see how this would be hard to mandate though.

Re polyfilling: I'd like to avoid this as my use-case for `ts-rest` is exporting the client as a separate library, and polyfilling a global in a library does not seem like a great idea to me.

If this still seems like a bad solution to you, would you be open to me contributing something to the docs that mentions this limitation? Mostly mentioning that if you want to use the `ts-rest` client for form uploads in Node you need to use `node 20` or polyfill the File global.

michaelangeloio:
> Thanks for your response!
> 
> I do get that you cannot set a filename to a Blob, but there are different ways of supplying filenames for e.g. an upload, e.g. as a separate form field. I do feel like it makes sense to allow Blob uploads, [since this is also supported in the browser](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append). I personally do not think `ts-rest` should "dictate" that only Files are allowed to be used in forms. (of course my actual reason is that it would be easier for me this way haha)
> 
> Maybe an API for adding filenames could be added, e.g. you could maybe pass a tuple of a blob and a string which which would be used in `formData.append`. I do see how this would be hard to mandate though.
> 
> Re polyfilling: I'd like to avoid this as my use-case for `ts-rest` is exporting the client as a separate library, and polyfilling a global in a library does not seem like a great idea to me.
> 
> If this still seems like a bad solution to you, would you be open to me contributing something to the docs that mentions this limitation? Mostly mentioning that if you want to use the `ts-rest` client for form uploads in Node you need to use `node 20` or polyfill the File global.

Hi! Is there a way we can meet in the middle perhaps? Maybe a way for us to add additional functionality to detect both?

tefkah:
@michaelangeloio  I'd be happy to meet in the middle!

You can still use a `Blob` and a filename to set the name of a `File` in a form like this in node 18

```ts
const form = new FormData()

form.set('file', new Blob(['hey']), 'hello.txt')

const file = form.get('file')

console.log(file.name) // hello.txt
```

Using this, you might be able to supply an API in `ts-test` like this maybe

```ts
const router = c.router({
 	upload: {
			// ...
			contentType: 'multipart/form-data',
			body: z.object({ 
				file: z.tuple([z.instanceof(Blob), z.string({description: 'Filename'})])
				// and/or if you are working in Node 20+ or the browser (or Deno/Bun or something)
				.or(z.instanceof(File)) 
 		}
})
```

Then you could upload things like
```ts
const { body } = await client.upload({file: [new Blob(['hey']), 'hello.txt']})
```
or
```ts
const { body } = await client.upload({ file: new File(['hey'], 'hello.txt')})
```

It might be easier to provide this as a helper, like

```ts
import { zodBlobFormInput, c } from '@ts-rest/core'

// const zodBlobFormInput = z.tuple([z.instanceof(Blob), z.string({description: 'Filename'})])
// typeof c.BlobFormInput = ['Blob', string] 

// ...
		file: zodBlobFormInput // or c.BlobFormInput
```

I'm not great at names but you know what I mean.

Then here, change this
https://github.com/ts-rest/ts-rest/blob/b9ef8d4a9c90c1be8f70ee6599ae1e6b71f6ab34/libs/ts-rest/core/src/lib/client.ts#L150-L156

to 
```ts
  Object.entries(body as Record<string, unknown>).forEach(([key, value]) => {
    if (Array.isArray(value)){
    	const { success, data, error } = zodBlobFormInput.safeParse(value)
    	if(!success){
    	 	// handle error
    	}
    	const [blob, filename] = data
    	formData.set(key, blob, filename)
    }
    // i'd still personally like to see this changed to `instanceof Blob` as it's more flexible and will not throw an error in node 18
    // but can concede this if you want to keep it
     else if (value instanceof File) {
      formData.append(key, value);
    } else {
      formData.append(key, JSON.stringify(value));
    }
  });
```


Thoughts?

Gabrola:
Closing as `File` has been available in Node since v18.13.0

PR #414, opened 2023-10-17
[OPEN-API] Pass through `contentType` to OpenApi schema

vancegillies:
Ran into this today, seems like this would be expected behaviour?
also updated the usage of `c.response` as its deprecated.


michaelangeloio:
@vancegillies  can you sync with main branch please?

michaelangeloio (approved):
LGTM

PR #419, opened 2023-10-20
fix: content-type text/html string body (#418)

### LINKED ISSUE #418: @ts-rest/client Content-type text/html returns blob body
Content-type "text/html" returns blob body, should return string body.

```js
import { z } from "zod";

export const loginApi = initContract().router({
  login: {
    method: "POST",
    path: "/login",
    body: z.object({
      email: z.string(),
      password: z.string(),
    }),
    responses: {
      200: z.string(),
    },
  },
});
```

The content-type is "text/html" when z.string() is the response. This does not match "text/plain" in client tsRestFetchApi().

dwidge:
Fixes #418.
Parse any "text/" content-type with text().

### DIFF
--- a/libs/ts-rest/core/src/lib/client.spec.ts
+++ b/libs/ts-rest/core/src/lib/client.spec.ts
@@ -368,7 +368,7 @@ describe('client', () => {
       expect(result.headers.get('Content-Length')).toBe('15');
     });
 
-    it('w/ a non json response (string)', async () => {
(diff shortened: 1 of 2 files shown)


oliverbutler (commented):
I'm sure this will work perfectly, but it'd be great to see a test case in the client.spec.ts for text/html as an example

Happy to make this test if you don't have an oppertunity

otherwise LGTM!

michaelangeloio (approved):

PR #424, opened 2023-11-02
Fix error due missing comma

miku86:
Properties in an object should be comma-separated, here one comma is missing, leading to an error.

I add a comma to fix the error.


oliverbutler (approved):
Thanks! LGTM

PR #445, opened 2023-12-07
feat(ts-rest/nest): http exception options in TsRestException

reecefenwick:
Adds support for an optional `HttpExceptionOptions` argument available on the base class `HttpException` from NestJS

Allows developers to include additional context i.e. a root cause error in `TsRestException`, which can be useful in exception filters to centrally catch and log exceptions

### DIFF
--- a/.changeset/metal-mugs-smoke.md
+++ b/.changeset/metal-mugs-smoke.md
@@ -2,4 +2,4 @@
 '@ts-rest/nest': minor
 ---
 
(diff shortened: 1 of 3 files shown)


michaelangeloio:
@reecefenwick LGTM, can you make an changeset with `npx changeset`?

reecefenwick:
Done @michaelangeloio 😀

michaelangeloio (approved):

Issue #450, opened 2023-12-16
@ts-rest/nest - throw TsRestException instead of mapping to a response?


reecefenwick:
Hi,

I am writing a http exception filter for my app to map all errors to a response, I was hoping that `TsRestExceptions` thrown would be caught in my exception filters, but alas it isn't

Looking at the code I found this here https://github.com/ts-rest/ts-rest/blob/main/libs/ts-rest/nest/src/lib/ts-rest-nest-handler.ts#L348-L353

Is there a need to map this response here?

Given `TsRestException` extends a `HttpException` from Nest, the same outcome would happen by nature of Nest (mapping to the response)

Happy to raise a PR, just wanted to sanity check, perhaps I'm missing something

PR #451, opened 2023-12-16
feat(ts-rest/nest): allow TsRestException to be handled in Nest exception filters

reecefenwick:
* Currently when a `TsRestExceptionFilter` is thrown from a controller, these cannot be caught in a nestjs exception filter. They are instead mapped directly to a response by the `TsRestHandlerInterceptor`
* This change now catches `TsRestExceptionFilter` and re-throws as a nestjs `HttpException`, mapping the response status/body from the original error and nesting it as a cause

This should be a backwards compatible change
* The body/status from a `HttpException` are mapped to the response by default with NestJS, which is what the ts-rest interceptor was doing originally
* We still maintain existing behaviour around response validation

Related issue - https://github.com/ts-rest/ts-rest/issues/450

### DIFF
--- a/.changeset/funny-taxis-brush.md
+++ b/.changeset/funny-taxis-brush.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/nest': minor
+---
+
+Allow TsRestException to be handled by NestJS exception filters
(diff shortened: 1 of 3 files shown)


michaelangeloio:
LGTM, can you add a changeset with `npx changeset` @reecefenwick ?

reecefenwick:
Done @michaelangeloio 😃

michaelangeloio (approved):

PR #454, opened 2023-12-24
feat(open-api): allow examples on OAS body and responses

kkolstad:
I'm looking for a way to provide multiple examples for body and response object ideally in the ts-rest contract itself. This solution does so, albeit in a less than ideal way, by adding a propert to `SchemaObject` which really belongs on the eventual `MediaTypeObject` one level up the OAS hierarchy.

This is a feature I'd really like to see. I'm happy to help with a better solution if this is something you'd like to support.

Screenshot 2023-12-23 at 10 28 52 PM


michaelangeloio (commented):
this LGTM, but can we get a unit test for it?

michaelangeloio:
also needs a changeset with `npx changeset`

Gabrola (approved):
Nice contribution! I agree it's less than ideal having to extend the `SchemaObject` but seems that there is no cleaner solution. I've renamed `examplesMap` to `mediaExamples` to clearly indicate that these are examples that will be added to the media type rather than the schema.

Issue #415, opened 2023-10-19
Migrate to TanStack Query v5


iamyuu:
Read the announcement post here: https://tanstack.com/blog/announcing-tanstack-query-v5
Read the migration guide here: https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5

techgerm:
This would be awesome! 🚀

Talent30:
Hi, @iamyuu @techgerm. I am trying to migrate it. Please join discord if you would like to help.

michaelangeloio:
One caveat to this: All of our packages are currently in sync. We'd have to break this convention to _not_  let others be forced to upgrade (should they want the latest @ts-rest/core package, for instance). 

If the community is OK with upgrading to react-query v5, I'm also cool with it, so as long as we properly communicate @ts-rest/vX will force you to upgrade react query.

PontusAxelsson:
It is peerDependency to "@tanstack/react-query": "^4.0.0", would be good if it could handle both like:
"peerDependencies": {
    "react": "^16.8.0 || ^17.0.0 || ^18.0.0",
    "@tanstack/react-query": "^4.0.0" || "^5.0.0",
    "zod": "^3.22.3"
  },

mxmnci:
+1 on this one!

AdhamMoussa:
+1

fruchtzwerg:
I suggest to do this incrementally. v4 already supports object syntax for query options (1st item in the migration guide):
https://github.com/ts-rest/ts-rest/blob/main/libs/ts-rest/react-query/src/lib/react-query.ts#L91
```
- useQuery(queryKey, dataFn, options)
+ useQuery({ queryKey, queryFn, ...options })
```

This could be done under the hood as a non-breaking change, thus supporting v4 and v5 as a peer dependency. I would imagine that this fixes 90% of use-cases.

rjay98:
@Gabrola, we're using `ts-rest` in production and loving it - thank you all for maintaining & developing this! 

Wanted to flag that `react-query` v5 is definitely a top priority for us, mostly as a way to start incorporating some of the new features (e.g., https://github.com/TanStack/query/pull/7481) which has powerful new tools for promise handling between client/server components.

ChiefORZ:
> One caveat to this: All of our packages are currently in sync. We'd have to break this convention to _not_ let others be forced to upgrade (should they want the latest @ts-rest/core package, for instance).
> 
[...]

PR #457, opened 2024-01-02
feat: add initial compatibility with react-query v5

fruchtzwerg:
relates to #415 

This PR establishes compatibility with TanStack Query v5 by using object-syntax internally. 
The external API of @ts-rest/react-query remains unchanged.
Compatibility with v4 remains unchanged.

It prevents React from throwing a runtime error on app launch.
It does **not** implement a full migration to v5 but lays the groundwork to get the ball rolling and apps running again.

Since this is not a fully featured migration I marked it as a patch.

### DIFF
--- a/.changeset/five-singers-jog.md
+++ b/.changeset/five-singers-jog.md
@@ -1,0 +1,7 @@
+---
+'@ts-rest/react-query': patch
+'@ts-rest/vue-query': patch
+---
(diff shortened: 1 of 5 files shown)


michaelangeloio (approved):
Regression tested on a live project. LGTM! 🚀

PR #461, opened 2024-01-06
feat(core): Support application/x-www-form-urlencoded content type

### LINKED ISSUE #453: Unable to use `content-type` header
It seems like the `content-type` header is being stripped from my request when passing through:

```typescript
const contract = c.router({
	getAccessToken: {
		method: "POST",
		path: "/oauth_token.do",
		body: z.string(),
		responses: {
			200: z.object({
				access_token: z.string(),
				token_type: z.string(),
				expires_in: z.number(),
				scope: z.string(),
				refresh_token: z.string(),
			}),
		},
		headers: z.object({
			"content-type": z.literal("application/x-www-form-urlencoded"),
		}),
	},
})
```

I understand there's a contentType attribute when creating the route. However, it only has strict options:

image

alex-way:
This pull request aims to support `application/x-www-form-urlencoded` `content-type`. Currently it's unsupported and attempting to use the correct headers causes issues. Closes #453

### DIFF
--- a/.changeset/thin-numbers-taste.md
+++ b/.changeset/thin-numbers-taste.md
@@ -2,4 +2,4 @@
 '@ts-rest/core': minor
 ---
 
-Add support for x-www-form-urlencoded content-type
+feat: `@ts-rest/core`: Add support for `x-www-form-urlencoded` content-type to core client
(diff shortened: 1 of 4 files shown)


michaelangeloio:
LGTM, But this needs a changeset. `npx changeset`

alex-way:
@michaelangeloio Added the changeset - Let me know if there's any issues 😄

michaelangeloio:
LGTM! Thank you

PR #469, opened 2024-01-10
docs: added pathParams examples to contract docs

### LINKED ISSUE #465: Documentation lacks `pathParams` details
There's no clear information in the docs on how to implement `pathParams` in a contract.  It should be clear that `pathParams` expects a zod schema - I think it'd make sense to include an example using `z.coerce.number()` as well. 

Happy to take a swing at updating the docs w/ some guidance on where it'd make sense to include this information. 

Something like:

```ts
  getUser: {
    method: 'GET', 
    path: '/users/:id', 
    pathParams: z.object({
      id: z.string()
    }), 
    // ...
  }
})

// or, if the ID is a number, use zod coerce:
  getUser: {
    method: 'GET', 
    path: '/users/:id', 
    pathParams: z.object({
      id: z.coerce.number()
    }), 
    // ...
  }
})
```

ihusemann:
aims to close #465

### DIFF
--- a/apps/docs/docs/core/core.md
+++ b/apps/docs/docs/core/core.md
@@ -1,10 +1,10 @@
 # Contract
 
-Define a contract with the `@ts-rest/core` package, you may nest routers within a router, generally you'd want a router for each nested resource e.g. `/users/:id/posts` could have a nested router `contract.users.posts`, this path is what you'd use on the client to query the API.
+Use the `@ts-rest/core` package to define a contract. Nesting routers can help organize your resources. For example, `/users/:id/posts` could have a nested router `contract.users.posts`. This is the path that you'd use on the client to query the API.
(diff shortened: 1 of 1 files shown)


apps/docs/docs/core/core.md
```diff
@@ -67,6 +67,38 @@ export const contract = c.router({
...
+    ...,
+    path: '/api/posts/:id',
+    pathParams: z.object({
+      id: z.string(),
+    }),
+  }
+});
+```
+
+Since URLs are just strings, any type other than `string` would need to be coerced or transformed into the required type. Using zod's `.coerce` for example:
```
michaelangeloio, line 87: love this, thank you

michaelangeloio (approved):

PR #473, opened 2024-01-16
feat: allow file downloads with express

### LINKED ISSUE #431: Explicit support for file download
Hello!

I think it would be nice for ts-rest to support file download. I've been able to achieve it doing like this:

```ts
const contract = initContract().router({
  downloadFile: {
    method: "POST",
    path: "/download-file",
    body: z.object({
      id: z.string(),
    }),
    responses: {
      200: z.unknown(),
    },
  },
});

function handler() {
  // ...
  return {
      status: 200,
      body: Buffer.from("ABC"),
      headers: {
        "Access-Control-Expose-Headers": "Content-Disposition",
        "Content-Type": "application/pdf",
        "Content-Disposition": `attachment; filename="file.pdf"`,
      },
    };
}
```

damieng57:
closes  #431


michaelangeloio:
Hey there! Looks good to me. What is the photo for, though? GitHub is listing the file as a Binary

michaelangeloio:
Also, needs a changeset! `npx changeset` or `npm i -g @changeset/cli && changeset`.

michaelangeloio:
Can you link the issue you're trying to close in the description?

damieng57:
Hi, @michaelangeloio and thank you for reviewing this PR. Now, it should be ok. However, I never used changeset/nx before, if I do something wrong, I will fix it again.

The picture is used in unit tests. I compare the original version with the version returned in the response. This is to be sure that it is indeed the same.

michaelangeloio (approved):
Thank you!

PR #482, opened 2024-02-01
Correctly add `style: deepObject` for optional objects

### LINKED ISSUE #477: getQueryParametersFromZod not checking inner type to determine isObject
I have a optional query parameter with the following definition:
```
z.object({
  page: z
    .object({
      limit: z.coerce.number().int(),
    })
    .optional(),
}),
```
When z.optional is specified, the zod object becomes a ZodOptional with the inner type as ZodObject. This means that the isObject flag is never true on this line: https://github.com/ts-rest/ts-rest/blob/4d5a7e93c69be854b040577b3d5d7b7b298462af/libs/ts-rest/open-api/src/lib/ts-rest-open-api.ts#L122

I require the style to be set to deepObject to allow my openapi doc generator tool I'm using to correctly show the examples.

I suspect it can be fixed by checking the inner type as follows.
```
const isZodObject = (obj) => {
    while (obj._def.innerType) {
        obj = obj._def.innerType;
    }

    return obj._def.typeName === 'ZodObject';
};
```

Alternatively, exposing the explode and style in the parameter definition can do the trick which I believe is less of a breaking change.

Is there any other options to get style set to DeebObject or will a fix need to be implemented?

robhung:
Fixes #477

### DIFF
--- a/libs/ts-rest/open-api/src/lib/ts-rest-open-api.spec.ts
+++ b/libs/ts-rest/open-api/src/lib/ts-rest-open-api.spec.ts
@@ -619,6 +619,75 @@ describe('ts-rest-open-api', () => {
       });
     });
 
+    it('works with zod optional query parameters', () => {
+      const routerWithRefine = c.router({
+        endpointWithZodRefine: {
+          method: 'GET',
+          path: '/optional',
+          responses: {
+            200: c.type<null>(),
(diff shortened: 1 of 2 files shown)


Gabrola (approved):
Nice one, thank you!

PR #483, opened 2024-02-03
Use consistent naming in documentation and show dependencies

ibash:
I haven't used ts-rest yet but when evaluating whether it would be a good fit I was confused by where `initServer` and `api` were coming from.

This commit renames `api` to `contract` for next.js to be consistent with the other library usages and adds a couple comments on where it comes from.

It also adds the import of initServer in the quick start so it's clear that's something ts-rest provides.

--

I haven't actually used ts-rest yet, so please let me know if this commit is wrong.


abdirahmn1:
i was also confused when reading the docs where everything was coming from, this was a big headache for me when getting started, but once you get the hang of it, it's VERY VERY smooth and enjoyable experience, i love it :+1:

Gabrola (approved):
Thank you!

PR #487, opened 2024-02-07
docs: add fake api guide

avancealaska:
This PR adds a new page to the "Guides" section, detailing how ts-rest can be used to create a fake API for testing purposes.


oliverbutler (approved):
Nice guide, this could be helpful for some folks, would be nicer if the lib had better support for this approach without using express of course!

oliverbutler:
@all-contributors please add @avancealaska  for docs

Issue #442, opened 2023-12-04
`useQuery` is expecting the `queryKey` again in the optional react-query params


hexcowboy:
with the example here (https://ts-rest.com/docs/react-query)

```ts
const queryResult = client.posts.get.useQuery(
  ['posts'], // <- queryKey
  { params: { id: '1' } }, // <- Query params, Params, Body etc (all typed)
  { staleTime: 1000 } // <- react-query options (optional)
);
```

there's a typescript compiler error:

```
Property 'queryKey' is missing in type '{ staleTime: number; }' but required in type 'UseQueryOptions<...>'.
```

right now the workaround is to just supply the `queryKey` again:

```ts
const queryKey = ['posts'];
const queryResult = client.posts.get.useQuery(
  queryKey,
  { params: { id: '1' } },
  { staleTime: 1000, queryKey } // <- supply it here as well and error goes away
);
```

hexcowboy:
> Do you mind making a PR for a docs? Or do we think this is a bug?

i'm kind of thinking this is a bug. why would the `queryKey` need to be supplied twice?

Gabrola:
`queryKey` is only required in the options in react-query v5. Make sure you have v4 instead as ts-rest is not compatible with v5 yet.

PR #489, opened 2024-02-11
omit queryKey from react-query options

AdhamMoussa:
#442

### DIFF
--- a/libs/ts-rest/react-query/src/lib/react-query.ts
+++ b/libs/ts-rest/react-query/src/lib/react-query.ts
@@ -84,7 +84,7 @@ const getRouteUseQuery = <
   return (
     queryKey: QueryKey,
     args?: ClientInferRequest<AppRouteMutation, ClientArgs>,
-    options?: TanStackUseQueryOptions<TAppRoute['responses']>,
+    options?: Omit<TanStackUseQueryOptions<TAppRoute['responses']>, 'queryKey'>,
   ) => {
     const dataFn = queryFn(route, clientArgs, args);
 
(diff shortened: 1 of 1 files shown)


Gabrola:
This will not fix anything as this change only changes the types for the internal API. The external facing types are here https://github.com/ts-rest/ts-rest/blob/main/libs/ts-rest/react-query/src/lib/types.ts

But I'm curious as to why you are getting this issue in the first place. `queryKey` is optional in the query options and I've never come across this issue before. @iceStorm can you make a quick reproducible repo or codesandbox?

iceStorm:
@Gabrola please kindly take a look at the following reproduction:

https://github.com/iceStorm/ts-rest-query-key-issue?tab=readme-ov-file#ts-restreact-query-querykey-issue

Gabrola:
@iceStorm the problem in your repo is that you have v5 of @tanstack/react-query installed which is incompatible with ts-rest. Simply installing v4 and adding it into your package.json will fix the problem. It seems that v5 was added to our valid peer dependencies in a recent PR, hence why it was installed automatically due to the fact that pnpm now installs peerDependencies by default instead of warning you. I will fix this now.

AdhamMoussa:
@Gabrola why do we need to stick with v4 instead of fixing the actual issue?
tbh this whole package for react-query is useless anyway. I don't see any value from it since the only thing it does is passing the arguments to the fetch function. it has poor TS types and a lot of issues. I ended up removing this package and just using react-query directly with ts-rest client. much better experience

Gabrola:
@AdhamMoussa because v5 is a breaking change. We cannot support both v4 and v5 at the same time, and since all of the ts-rest package versions are synchronized we'd have to release a major version release for all @ts-rest/* packages. We'd like to keep it the same way as it removes any complexity of compatibility between the different packages. Tackstack query v4 is still the most downloaded version until now, so there's no rush to upgrade. We also have other features and changes planned for ts-rest v4, so we'd rather have to release only one breaking major version release.

AdhamMoussa:
@Gabrola makes sense. Thanks

PR #492, opened 2024-02-20
refactor(nest): allow nestjs handlers to return a promise or a value

SharkEzz:
Hi !

This PR allow a NestJS handler to return either a Promise or a simple value (as all handlers may not be async functions).

This is my first contribution to this project, so feel free to tell me if something is missing :grinning:


Gabrola (approved):
Nice one, thank you!

LiamMorrow:
Hmm, this seems to have broken our typing for our async handlers.  We had a utility type which allowed us to give a return type to our handler methods.  This gave us much better type errors than when the handlers did not have annotated return types:
```

export type Handler<T extends AppRoute | AppRouter> = ReturnType<typeof tsRestHandler<T>>

export type HandlerResponse<T extends AppRoute> = ReturnType<Handler<T>>
```

Since this change, our async handlers can no longer be annotated with this type as it now includes a non promise type

image

Do you guys have any suggestions on how we can annotate our handler methods with a return type?

Gabrola:
@LiamMorrow you are right, will revert this change. Thank you!

PR #503, opened 2024-03-10
Adds support for creating single url routes in Next.js

dodiego:
Solves https://github.com/ts-rest/ts-rest/issues/504


Gabrola (changes requested):
Good contribution, thank you! Just needs a rebase/merge in main, and address the comments.

libs/ts-rest/next/src/lib/ts-rest-next.ts
```diff
@@ -309,4 +371,4 @@ export const createNextRouter = <T extends AppRouter>(
       throw e;
     }
   };
-};
```
dodiego, line 312: automatic prettier / eslint fix

dodiego:
Comments addressed and changes are synced with latest version of `main` branch. There is just one semicolon removal at the end of the file `libs/ts-rest/next/src/lib/ts-rest-next.ts` which was done automatically by prettier / eslint.

libs/ts-rest/next/src/lib/ts-rest-next.ts
```diff
@@ -186,42 +197,96 @@ export const createNextRoute = <T extends AppRouter>(
...
+  return handlerFactory((req) => {
+    const route = { ...appRoute, implementation: implementationHandler };
+    const pathParams = req.query as Record<string, string>;
+    const query = req.query;
+
+    const isValidRoute = isRouteCorrect(
+      appRoute,
+      req.url!.split('/').slice(1),
+      req.method as string,
+    );
```
Gabrola, line 252: Although the current code should work in most cases. The passed `pathParams` will be wrong if the route file/dir name `[foo]` does not match the name in the contract, despite `isValidRoute` being true due to using the actual URL rather than `req.query`. 

I think it'll also be better to also filter out the `pathParams` from `req.query` before passing it into `query`, but it won't really work if the names in the original `req.query` and `pathParams` don't match, but there's nothing we could really do at this point.

```suggestion
    const urlChunks = req.url!.split('/').slice(1);
    const pathParams = getPathParamsFromArray(urlChunks, route);
    const query = req.query
      ? Object.fromEntries(
          Object.entries(req.query).filter(
            ([key]) => pathParams[key] === undefined,
          ),
        )
      : {};

    const isValidRoute = isRouteCorrect(
      appRoute,
      urlChunks,
      req.method as string,
    );
```

Gabrola (changes requested):

Gabrola (approved):
Looks great, thank you! Great contribution 🙌

PR #522, opened 2024-03-22
feat: ts-rest client throw on error status code

### LINKED ISSUE #520: Feature Request: Create new ts-rest client that throws exceptions for error responses
We are currently utilizing the `ts-rest` library on both the client and server side of our application. The client side is built with React and uses `react-query`. However, we are not using the `@ts-rest/react-query` client due to compatibility issues with our usage patterns.

Our previous experience with `Zodios` and `Axios` for executing requests has led us to observe that the resolution of all responses (including non-success ones) can result in weird code flow. This issue is also observable in the provided examples.

### Example 

```ts
import { useQuery } from "@tanstack/react-query";

function App() {
  const { mutation, isPending, isError, isSuccess } = useMutation({
    mutationFn: (newPost) => client.addPost(newPost),
    onSuccess: (response) => {
        if (response.status === 200) // do success
        if (response.status === 500) // do error stuff
    },
    onError: (error) => {
        // do other error stuff
    }
  });

  return <div>
    <h1>Posts</h1>
    <div>{/** render posts */}</div>
    <button onClick={mutation}>Save post</button>
  </div>
}
```

The above example presents several issues:
- Upon execution of the mutation, all responses are directed to the `onSuccess` callback. This includes error paths, which is not ideal. Error-handling should not be performed in the `mutationFn`, as it results in a non-ideal `response` type (if we return `undefined` in the `mutationFn`, the type of `response` will be `Response | undefined`).
- The `onError` callback handles errors for the second time.
- The `isError`/`isSuccess` flags are incorrect, as an error response can be considered a success.
- A similar issue exists for error response handling in client router data loaders. The data loaders also have error handler callbacks, and in most cases, throwing the error makes error handling easier.

### Proposal

To address these issues, we propose the following:
- Develop a client that treats responses in the 2xx range as successful and resolves the promise. In most cases, there will be only a single 2xx response definition in the contract, simplifying success handling (you don't have to handle the error case to handle the success case).
- Responses outside of the success range should be thrown as errors. Errors should be thrown with an error object containing the error response.
- Implement a type-guard method that checks if a given error is thrown by a particular endpoint, `isTsRestError(route: AppRoute | AppRouter, status : ErrorHttpStatus, error: Error) : error is ...`
- (Optional) Create a configuration that marks particular responses as success or error. This could potentially be achieved via the contract definition (`{ status: 200, body: ...}, { status: 201, isError: true, body: ... }`). This could be a router configuration or client configuration similar to the `axios.validateStatus` function: [axios-http.com/docs/handling_errors](https://axios-http.com/docs/handling_errors)

Example from above with 
```ts
import { useQuery } from "@tanstack/react-query";
function App() {
  const { mutation, isPending, isError, isSuccess } = useMutation({
    mutationFn: (newPost) => client.addPost(newPost),
    onSuccess: (response) => {
       // response.status is only 200
       // do optimistic update, do success
    },
    onError: (error) => {
        if (isTsError(client.addPost.route, 500, error)) // do error stuff
        // do other error stuff
    }
  });

  return <div>
    <h1>Posts</h1>
    <div>{/** render posts */}</div>
    <button onClick={mutation}>Save post</button>
  </div>
}
```
- The success path is very clear, it will be response.body in most cases because there will be only one success status
- The error path is clear, all kinds of error are handlded in a single place. The type safe response remain.
- The `isError`/`isSuccess` flags are correct
- Error can be handled anywhere in the application in type safe manner, it can even be handled in the root error handler.

toteto:
Resolves #520

There are no breaking changes with this new addition. In order to get this behavior of the client to throw on error status code, a parameter of `throwOnErrorStatus: true` needs to be provided to the `initClient` options.

Open for changes or feedback to the code. If you don't see this as an addition of the `@ts-rest/core` client, looking for suggestion how such behavior can be achieved by separate client that can be in separate package/repository.


libs/ts-rest/core/src/lib/error-status-error.ts
```diff
@@ -0,0 +1,54 @@
...
+import { ErrorHttpStatusCode } from './status-codes';
+
+export class ErrorStatusResponseError<
+  TAppRoute extends AppRoute,
+  TErrorStatus extends ErrorHttpStatusCode = ErrorHttpStatusCode,
+> extends Error {
+  constructor(
+    public readonly route: TAppRoute,
+    public readonly response: ErrorResponse<TAppRoute, TErrorStatus>,
+    public readonly request: unknown,
```
Gabrola, line 12: Can't this be `ClientInferRequest<TAppRoute, ClientArgs>`?

libs/ts-rest/core/src/lib/error-status-error.ts
```diff
@@ -0,0 +1,54 @@
...
+> {
+  if (error == null) return false;
+  if (!(error instanceof ErrorStatusResponseError)) return false;
+
+  const errorStatusResponseError = error as ErrorStatusResponseError<
+    AppRoute,
+    ErrorHttpStatusCode
+  >;
+
+  if (errorStatusResponseError.route.path !== route.path) return false;
```
Gabrola, line 48: Need to check the method too, not only the  path.

libs/ts-rest/core/src/lib/error-status-error.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  it('should check for array of error code statuses', () => {
+    const isHealthError = isErrorStatusResponse(
+      unauthorizedHealthError,
+      router.health,
+      [401, 403, 500],
+    );
+
+    if (!isHealthError) fail('should be health error');
```
Gabrola, line 78: Use `expect(isHealthError).toBe(true)`

libs/ts-rest/core/src/lib/error-status-error.ts
```diff
@@ -0,0 +1,54 @@
...
+export function isErrorStatusResponse<
+  TAppRoute extends AppRoute,
+  TErrorStatus extends ErrorHttpStatusCode = ErrorHttpStatusCode,
+>(
+  error: unknown,
+  route: TAppRoute,
+  status?: TErrorStatus | TErrorStatus[],
+): error is ErrorStatusResponseError<
+  TAppRoute,
+  TErrorStatus extends undefined ? ErrorHttpStatusCode : TErrorStatus
```
Gabrola, line 38: `TErrorStatus extends undefined` will never be true because you already declared it as `TErrorStatus extends ErrorHttpStatusCode`. Even if not, `T extends undefined` in general doesn't work that way for optional parameters, because they resolve as `unknown` not `undefined`, and actually because `extends` is distributive you need to do `unknown extends T` not `T extends unknown`

This should work fine if you just replace `TErrorStatus extends undefined ? ErrorHttpStatusCode : TErrorStatus` with `TErrorStatus` because you already gave it a default value of `ErrorHttpStatusCode`, so it will take that if you don't pass `status`

Gabrola (changes requested):

toteto:
Closing in favour of having un-opinionated implementation

PR #532, opened 2024-03-27
feat: create helper functions for route and implementation separation

### LINKED ISSUE #521: Split a router in multiple files or variables
Hi,

I'm unable to split a router into multiple files or variables. I prefer not to create nested routers.
I'm encountering a TypeScript error because `getPost` is not recognized.

```javascript
const s = initServer();
const getPost = async ({ params: { id } }) => {
    return {
        status: 200,
        body: {
            id: "1",
            title: "my title",
            body: "000"
        }
    };
};
const createPost = async ({ body }) => {
    return {
        status: 201,
        body: {
            id: "1",
            title: "my title",
            body: "000"
        }
    };
};
    getPost,
    createPost
});
createExpressEndpoints(contract, router, app);
```

This code is fine, but I prefer not to have multiple endpoints in a single file.

```javascript
    const s = initServer();

        // move getPost in routers/post/get.ts
        getPost: async ({ params: { id } }) => {
            console.log(id);
            return {
                status: 200,
                body: {
                    id: "1",
                    title: "my title",
                    body: "000"
                }
            };
        },
        // move getPost in routers/post/create.ts
        createPost: async ({ body }) => {
            return {
                status: 201,
                body: {
                    id: "1",
                    title: "my title",
                    body: "000"
                }
            };
        }
    });

    createExpressEndpoints(contract, router, app);
```

toteto:
Motivated by few issues (like #521) and comments on Discord, I am adding few methods that will aid into easier creation of reusable endpoints/responses/implementations.

If the changes in `ts-res/express` are desired, I am willing to try to port them to other server packages.

Resolves #521

### DIFF
--- a/libs/ts-rest/core/src/lib/helpers.ts
+++ b/libs/ts-rest/core/src/lib/helpers.ts
@@ -1,0 +1,36 @@
+import {
+  type AppRoute,
+  type ContractAnyType,
+  type ContractNullType,
+  type ContractOtherResponse,
+  type ContractPlainType,
+  ContractPlainTypeRuntimeSymbol,
+} from './dsl';
+
+/**
+ * Create a type that would be used in contract. Valid only at compile time. For having runtime safety, use `zod` types.
+ */
+export const makeType: <T>() => T extends null
+  ? ContractNullType
+  : ContractPlainType<T> = ContractPlainTypeRuntimeSymbol;
+
+/**
+ * Create route definition that would be used in contract.
+ */
+export function makeRoute<const TAppRoute extends AppRoute>(
+  route: TAppRoute,
+): TAppRoute {
+  return route;
+}
+
+/**
+ * Create a response definition that would be used for route definition in a contract.
+ */
+export function makeResponses<
+  const TResponses extends Record<
+    number,
+    ContractAnyType | ContractOtherResponse<ContractAnyType>
+  >,
+>(responses: TResponses): TResponses {
+  return responses;
+}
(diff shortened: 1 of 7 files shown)


libs/ts-rest/core/src/lib/helpers.ts
```diff
@@ -0,0 +1,36 @@
...
+  type ContractPlainType,
+  ContractPlainTypeRuntimeSymbol,
+} from './dsl';
+
+/**
+ * Create a type that would be used in contract. Valid only at compile time. For having runtime safety, use `zod` types.
+ */
+export const makeType: <T>() => T extends null
+  ? ContractNullType
+  : ContractPlainType<T> = ContractPlainTypeRuntimeSymbol;
```
toteto, line 15: > [!NOTE]  
> `makeType` is born out of the need to create a type without having `c.type` when constructing with `makeRoute`/`makeResponses`. 
> Any reasoning to having such stateless utility function behind object `c`? Can we get this standalone utility?

Gabrola:
There's no need for the core library helpers since they all exist in the `initContract()` type helper. `makeRoute` is covered by `c.query()` and `c.mutation()`

For now we'd prefer to stick to the `const c = initContract()` convention

The express helper can be achieved much easier, so I'll just make a quick to change to cover this need

toteto:
Having to import `initContract` or `initServer` just to use stateless utility methods seems like extra step and extra computation for anyone that is not 100% aware of the source code. Seems like the resulting type of `c.mutation/query` is not portable as `makeMutation` or `makeQuery` would be. This is my opinion at least. 

Any opinion on the `makeResponses` utility?

Gabrola:
We will be exporting `c` or similar directly in the next major version, and the old stateless method will be removed. However, exporting the utilities separately, not contained in an object, makes discovery harder and dependency on the docs even more - think of zod's `z`

PR #534, opened 2024-03-28
fix(core): ts-rest overwriting a supplied content-type header

### LINKED ISSUE #531: content-type header is overwritten
since the update to version **0.37.0** the **content-type** header is overwritten on all requests:

```ts
const updateUser = userClient.patchUser.useMutation();

updateUser.mutate(
  {
    body: data,
    headers: {
      "content-type": "application/merge-patch+json",
    },
    params: {
      id: user?.id,
    },
  }
)
```

We are overriding the content-type header on PATCH requests...
In the network tab we will always see `"content-type": "application/json"`

This used to work with version **0.36.0** without a problem.

In the source i see this block, which can be the cause of the problem:

```ts
// /libs/ts-rest/core/src/lib/client.ts#L251
} else if (body !== null && body !== undefined) {
  fetcherArgs = {
    ...fetcherArgs,
    contentType: 'application/json',
    headers: {
      ...fetcherArgs.headers,
      'content-type': 'application/json',
    },
    body: JSON.stringify(body),
  };
}
```

Gabrola:
Fixes #531

### DIFF
--- a/.changeset/lovely-snails-divide.md
+++ b/.changeset/lovely-snails-divide.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/core': patch
+---
+
+Fix ts-rest overwriting a supplied content-type header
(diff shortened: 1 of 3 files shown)


ChiefORZ (approved):
loooks good 🙏

PR #538, opened 2024-03-28
fix: validate response

toteto:


libs/ts-rest/core/src/lib/client.ts
```diff
@@ -215,6 +215,7 @@ export const fetchApi = ({
     rawBody: body,
     rawQuery: query,
     contentType: undefined,
+    validateResponse: clientArgs.validateResponse,
```
toteto, line 218: @Gabrola maybe check if there are other parameters that are forgotten?

apps/docs/docs/core/core.md
```diff
@@ -346,3 +346,32 @@ const parentContract = c.router(
   }
 );
 ```
+
+### Validate Response
```
Gabrola, line 350: This should be part of the `core/fetch` page instead

.changeset/beige-forks-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ts-rest/core': minor
```
Gabrola, line 2: Patch release

.changeset/beige-forks-sit.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+fix client not validating response
```
Gabrola, line 5: We capitalize these because they are part of the changelog

Gabrola (changes requested):
Looks good, just some minor comments

apps/docs/docs/core/core.md
toteto: Done b8fa3e8

.changeset/beige-forks-sit.md
toteto: Done 0d0af4f

.changeset/beige-forks-sit.md
toteto: Done 0d0af4f

Gabrola (approved):
LGTM! Just made some small changes to the docs

PR #539, opened 2024-03-28
feat: c.responses utility

toteto:


libs/ts-rest/core/src/lib/dsl.ts
```diff
@@ -196,12 +196,20 @@ type ContractInstance = {
    * A single query route, should exist within
    * a {@link AppRouter}
    */
-  query: <T extends AppRouteQuery>(query: T) => T;
+  query: <const T extends AppRouteQuery>(query: T) => T;
```
toteto, line 199: Note: The addition of `const` allows for better type inferring when using contract that has such route in `initContract`. Without it, the client type is incorrectly inferred.

Gabrola: Unfortunately, this is a breaking change because const generics are part of typescript 5.0

Gabrola: Unfortunately, this is a breaking change because const generics are a part of typescript 5.0. Users will have to use `as const` in this case.

Gabrola: Not that TS follows semver anyway, but it's still a breaking change. Although we use TS 5.2 internally, we are very careful not to use newer TS features in a way that would break our own semver in the emitted type library type files

toteto: Removed the `const` type in 1616ba3.

Can you think of an alternative to this feature? Was trying with `Narrow<T> but the resulting type ended up being:

```ts
type ct = {
    getPost: {
        readonly method: "GET";
        readonly responses: {
            readonly 200: {
                id: number;
                readonly [tag]: "ContractPlainType";
            };
        };
        path: "/posts/:id";
    };
}
```

Gabrola: I think `Narrow` can be changed to work, but the problem is that it is a recursive type and some people have already been hitting some TS limits with type instantiation, so I think it's just better to use `as const` for now. We're already forced to do that with the `metadata` part of the contract, and in Nest controllers for the response status code, so it's not a totally new problem for us.

PR #543, opened 2024-03-29
fix: empty response

### LINKED ISSUE #542: Having `void` responses causes response parsing to fail
When there is contract that expect empty body (`z.void()` or `c.type<void>()`) and the server doesn't return body (it returns `body: undefined`) the parsing of the response with `result.json()` throws error `Error: Unexpected end of JSON input`.

Repro: https://stackblitz.com/edit/stackblitz-starters-8pivcu?file=index.js

toteto:
Resolves #542

Open to ideas to better handle this. Current thinking is:
- Check for `content-length` header, if there is such and the length is `0`, there is no need to try to parse (nor get) the body. 
- Instead of directly parsing to `json`, first get the `text` value and if the value is non-empty, JSON parse it.

Open to ideas to write unit test for this case. Because the `client` tests rely on mocked `fetch`, testing is not possible this way.

### DIFF
--- a/.changeset/hungry-masks-carry.md
+++ b/.changeset/hungry-masks-carry.md
@@ -1,0 +1,6 @@
+---
+'@ts-rest/core': minor
+'@ts-rest/next': minor
+---
+
+fix empty result throwing parse error
--- a/libs/ts-rest/core/src/lib/client.spec.ts
+++ b/libs/ts-rest/core/src/lib/client.spec.ts
@@ -135,6 +135,13 @@ export const router = c.router(
         200: c.type<{ message: string }>(),
       },
     },
+    emptyResponse: {
+      method: 'GET',
+      path: '/empty',
+      responses: {
+        200: c.type<void>(),
+      },
+    },
     upload: {
       method: 'POST',
       path: '/upload',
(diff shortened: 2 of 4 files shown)


Gabrola:
This really doesn't sit right with me, as technically an empty-response should not have a `content-type` of json. Technically this is a malformed response. We'll need to handle this correctly on the server first.
One more thing, I believe expressjs strips the `content-length` header for 204 responses, so this looks like a fragile solution. Let me think of something

toteto:
> This really doesn't sit right with me, as technically an empty-response should not have a `content-type` of json. Technically this is a malformed response. We'll need to handle this correctly on the server first.

Agree that solving this in the express lib would be first fix. But would also say that if empty response is expected, the client should be able to handle it regardless of the headers. In the handling I mean also throw validation error if the response ends up not being empty.

> One more thing, I believe expressjs strips the `content-length` header for 204 responses, so this looks like a fragile solution. Let me think of something

If you look into the response from the example I have attached in the issue, there is no content-length header even though the response status is `200 OK`. So would say that the content-length is striped when there is no content. This is why I have the header as conditional

badeleux:
I think this is still an issue. While I agree that it’s a server-side problem, unfortunately, it’s not always possible to expect the server to be changed.

If the server returns `204 No Content`, empty body, and `content-type: application/json`

PR #545, opened 2024-03-30
feat: support HEAD requests

### LINKED ISSUE #541: Feature request: HEAD requests
It seems that currently, according to these types, the AppRoute is currently not supporting `HEAD` requests.

https://github.com/ts-rest/ts-rest/blob/56b955f82118e251a13b1e8766f25ca02e91356a/libs/ts-rest/core/src/lib/dsl.ts#L48-L66

They act pretty much like a GET request, but do not provide a body in the response. This can be useful sometimes, when only want to know the status of the request, or something related to the headers, without having to download the whole content of the response.

Is there any way to handle these in this library? I would also make use of the react-query wrapper, I don't know if something is needed here as well, of it inherits it from the core package.

julienmonnard:
Resolves #541.

As this is my very first PR in this project, feel free to let me know anything that I might have done wrong, or that I missed.

I fixed some tests that were affected by my changes, but as for new tests, my knowledge of the repo is still quite limited, so not sure what would make sense to add, and where.

### DIFF
--- a/apps/docs/docs/core/fetch.md
+++ b/apps/docs/docs/core/fetch.md
@@ -15,7 +15,7 @@ To customise the fetching behaviour, add an `api` function, please see [Custom A
 
 Your contract uses regular HTTP methods to define the type of request you want to make. We use this type to infer whether you are doing a `query` or a `mutate` request.
 
-Any requests that use the `GET` method will be treated as a `query` request, and any requests that use the `POST`, `PUT`, `PATCH` or `DELETE` methods will be treated as a `mutate` request.
+Any requests that use the `GET` or the `HEAD` method will be treated as a `query` request, and any requests that use the `POST`, `PUT`, `PATCH` or `DELETE` methods will be treated as a `mutate` request.
 
 ```typescript
 const { data } = await client.posts.get();
@@ -136,9 +136,7 @@ Objects implementing `.toJSON()` will irreversibly be converted to JSON, so you 
 For example, Date objects will be converted ISO strings by default, so you could handle this case like so:
 
 ```typescript
-const dateSchema = z
-  .union([z.string().datetime(), z.date()])
-  .transform((date) => (typeof date === 'string' ? new Date(date) : date));
+const dateSchema = z.union([z.string().datetime(), z.date()]).transform((date) => (typeof date === 'string' ? new Date(date) : date));
 ```
 
 This will ensure that you could pass Date objects in your client queries. They will be converted to ISO strings in the JSON-encoded URL query string, and then converted back to Date objects on the server by zod's parser.
--- a/libs/ts-rest/core/src/lib/client.ts
+++ b/libs/ts-rest/core/src/lib/client.ts
@@ -1,4 +1,10 @@
-import { AppRoute, AppRouteMutation, AppRouter, isAppRoute } from './dsl';
+import {
+  AppRoute,
+  AppRouteMutation,
+  AppRouteQuery,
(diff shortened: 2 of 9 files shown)


julienmonnard:
I see now the discussion on #543, and I think it's also somewhat applicable here, actually, because by definition the response of a HEAD request should be empty.
Based on that, I don't think the specific handling I added in the core client is the right approach, so I will keep an eye on that other discussion...

Gabrola:
I think the discussion you're referring to is not that much of a big deal in this case, to be honest. The code looks good initially, however, I could see a major usability issue though; almost always the `HEAD` and `GET` methods point to the exact same endpoint. In fact, express (and as a result Nestjs using the express adapter), will automatically handle `HEAD` requests for any `GET` routes defined. So by defining a separate `HEAD` endpoint in the contract would force someone to implement a separate route handler for that endpoint.

You would also expect the contract endpoint definitions to be exactly the same minus the response bodies, so it would be a cumbersome and error-prone having to duplicate the endpoint, and making sure all future changes are duplicated exactly.

I think ideally we should be able to define the endpoint with a method of `'GET' | 'HEAD'`, but it would still be a bit complex when dealing with the intricacies of the separate frameworks.

Gabrola:
I will close this for now, but feel free to discuss this further here or re-open if you come up with a solution

julienmonnard:
Right... Damn, I thought I would pick this one because it would be relatively straightforward, but it is more tricky than it seems in the first place 😛 
I think I got affected by the fact that in our project, we have only a handful of `HEAD` requests, and they are not matching the `GET` version of the same endpoint. So it made sense to me that we would add a separate endpoint in the contract, and a separate handler as well.
But it does make sense that most of the time, it should match the associated `GET`.

I admit that I am not very experienced with the server side of things. Our project has a separate backend, written in Java, and managed by a different team.

So if I understood correctly what you're saying is that basically, the current version of my PR would actually offer a new feature to most users, with a "not-so-ideal" usability. But for express, it would actually be a regression because it would break the way `HEAD` requests are automatically handled now?

Gabrola:
@julienmonnard yes, that is correct. In addition, I think the case where you have mismatching HEAD and GET requests is super edge case, that would not be needed by many (or any) people.

Anyway, I can rest easy that this is not a blocker for you, and you can workaround this limitation by using a custom client with custom args.

```typescript
const client = initClient(contract, {
  ...
  api: async (args: ApiFetcherArgs & { isHeadRequest?: boolean }) => {
    return tsRestFetchApi({
      ...args,
      ...(args.isHeadRequest && { method: 'HEAD' }),
    });
  },
});
```

Obviously, your response type will incorrectly indicate the existence of a response body, but I think you easily ignore that :)

Gabrola:
@julienmonnard yes, that is correct. In addition, I think the case where you have mismatching HEAD and GET endpoints is super edge case, that would not be needed by many (or any) people.

Anyway, I can rest easy that this is not a blocker for you, and you can workaround this limitation by using a custom client with custom args.

```typescript
const client = initClient(contract, {
  ...
  api: async (args: ApiFetcherArgs & { isHeadRequest?: boolean }) => {
    return tsRestFetchApi({
      ...args,
      ...(args.isHeadRequest && { method: 'HEAD' }),
    });
  },
});
```

Obviously, your response type will incorrectly indicate the existence of a response body, but I think you easily ignore that :)

julienmonnard:
Alright then, I will give it a try like this. Thanks for the feedback 🙂

PR #581, opened 2024-04-14
feat (ts-rest/core): allow merging router metadata to route metadata

iceStorm:
In our project, we need a mechanism to set metadata for a whole router (and for a sub-router, of course).

Then routes inside a router can inherit this metadata.

Use case:
- A router sets metadata with:
```
{
   requireAuth: true
}
```
- Then all routes inside the router will have `requireAuth: true` along with their own metadata.

### DIFF
--- a/libs/ts-rest/core/src/lib/dsl.ts
+++ b/libs/ts-rest/core/src/lib/dsl.ts
@@ -145,6 +145,7 @@ type ApplyOptions<
     responses: 'commonResponses' extends keyof TOptions
       ? Prettify<Merge<TOptions['commonResponses'], TRoute['responses']>>
       : TRoute['responses'];
+    metadata: TOptions['metadata'];
   }>;
 
 /**
@@ -173,6 +174,7 @@ export type RouterOptions<TPrefix extends string = string> = {
    * @deprecated Use `validateResponse` on the client options
    */
   validateResponseOnClient?: boolean;
+  metadata?: unknown;
 };
 
 /**
@@ -273,6 +275,12 @@ const recursivelyApplyOptions = <T extends AppRouter>(
               ...options?.commonResponses,
               ...value.responses,
             },
+            metadata: options?.metadata
+              ? {
+                  ...options?.metadata,
+                  ...(value.metadata ?? {}),
+                }
+              : value.metadata,
           },
         ];
       } else {
--- a/libs/ts-rest/core/src/lib/dsl.ts
+++ b/libs/ts-rest/core/src/lib/dsl.ts
@@ -145,7 +145,9 @@ type ApplyOptions<
     responses: 'commonResponses' extends keyof TOptions
       ? Prettify<Merge<TOptions['commonResponses'], TRoute['responses']>>
       : TRoute['responses'];
-    metadata: TOptions['metadata'];
+    metadata: 'metadata' extends keyof TOptions
+      ? Prettify<Merge<TOptions['metadata'], TRoute['metadata']>>
+      : TRoute['metadata'];
   }>;
 
 /**
@@ -169,12 +171,12 @@ export type RouterOptions<TPrefix extends string = string> = {
   strictStatusCodes?: boolean;
   pathPrefix?: TPrefix;
   commonResponses?: Record<number, AppRouteResponse>;
+  metadata?: unknown;
 
   /**
    * @deprecated Use `validateResponse` on the client options
    */
   validateResponseOnClient?: boolean;
-  metadata?: unknown;
 };
 
 /**
--- a/.changeset/twelve-ghosts-eat.md
+++ b/.changeset/twelve-ghosts-eat.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/core': minor
+---
+
+Add router-level metadata
--- a/libs/ts-rest/core/src/lib/dsl.spec.ts
+++ b/libs/ts-rest/core/src/lib/dsl.spec.ts
@@ -615,6 +615,46 @@ describe('contract', () => {
     >;
   });
 
+  it('should merge metadata options from router to its routes', () => {
+    const contract = c.router(
+      {
+        getPost: {
+          method: 'GET',
+          path: '/posts/:id',
+          responses: {
+            200: c.type<{ id: number }>(),
+          },
+        },
+        deletePost: {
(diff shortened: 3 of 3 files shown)


iceStorm:
@Gabrola Hii, could you help to assess this enhancement 😃

iceStorm:
Also, we need to access to a route's metadata inside the custom API fetcher callback. Like if `requireAuth: true` then we will attach an Authorization header to a request.

[Accessing the raw body and content type](https://ts-rest.com/docs/core/custom#accessing-the-raw-body-and-content-type)

Could you help advice some ideas...

libs/ts-rest/core/src/lib/dsl.ts
```diff
@@ -236,10 +237,11 @@ const recursivelyApplyOptions = <T extends AppRouter>(
               ? options.pathPrefix + value.path
               : value.path,
             headers: zodMerge(options?.baseHeaders, value.headers),
+            metadata: zodMerge(options?.metadata, value.metadata),
```
Gabrola, line 240: We don't want to inadvertently set `metadata` to an empty object, if it is not set for in either router options or the route. Also no need to use `zodMerge` as we are not merging schemas.

```suggestion
            metadata: options?.metadata
              ? {
                  ...options?.metadata,
                  ...value.metadata,
                }
              : value.metadata,
```

Gabrola (changes requested):
Thank you for the contribution! Can you please add a test to `dsl.spec.ts`?
Also your branch is quite a bit behind and is conflicting, can you fix that too?

Gabrola:
> Also, we need to access to a route's metadata inside the custom API fetcher callback. Like if a route has metadata with `requireAuth: true` then we will attach an Authorization header to the request.
> 
> [Accessing the raw body and content type](https://ts-rest.com/docs/core/custom#accessing-the-raw-body-and-content-type)
> 
> Could you help advice some ideas... @Gabrola

`route` is passed to the API fetcher function

libs/ts-rest/core/src/lib/dsl.ts
iceStorm: Got it. I didn't know about how the `zodMerge` function works. Also thought about having undefined if both are unset ;))

iceStorm:
> Thank you for the contribution! Can you please add a test to `dsl.spec.ts`? Also your branch is quite a bit behind and is conflicting, can you fix that too?

yess, flying on it 😃

Gabrola (changes requested):
You also need to merge the types in `ApplyOptions`

libs/ts-rest/core/src/lib/dsl.ts
iceStorm: Ahh, since `metadata` can be any type, it's unsafe to make an object and spread like this, isn't it...

iceStorm:
> You also need to merge the types in `ApplyOptions`

Updated. I'm not sure if it's aligned or not; it's my first time contributing to OSS. Kindly help to check again 😃

libs/ts-rest/core/src/lib/dsl.ts
iceStorm: It's hard to tell when a route's metadata is a string and its router's metadata is an object. The resulting type would be an intersection between the two, but I'm not sure how it would behave in real-life scenarios...

image

iceStorm: IMO, in most cases, the `metadata` should always be an object for consistency... 😃

iceStorm: In that case, the only changes are like below, if you are comfortable 😃

image

image

Gabrola: We can just let it be so it doesn't break any existing code. Nothing breaks as long as router-level metadata are used. It will only break if someone adds router-level metadata, and either of them are not objects, but it should be expected by the user that this breaks it.

Gabrola:
@iceStorm I've just fixed the type merging and add some type tests. Should be good to go. Thank you!

iceStorm:
> @iceStorm I've just fixed the type merging and add some type tests. Should be good to go. Thank you!

Wow, so fast 😃

### Also, sorry to bother you again. May I consult if there is any way to change the type of `metadata` globally for a project? Like using `declare global`, `declare module`...

Let's say we have a common interface for metadata:
```TypeScript
interface ApiRouteMetadata {
   requireAuth?: boolean;
   ...
}
```

And we want the interface to automatically be inferred when setting metadata for a route or a router. For consistency.

### The current idea that pop out of my head is to create a function:
```TypeScript
import { AppRoute as TsRestRoute } from '@ts-rest/core';

export type ApiRouteMetadata {
    requireAuth?: boolean;
}

export const SetApiRouteMetadata = (metadata: ApiRouteMetadata) => metadata;
```

### Then call the function when defining a contract:
```TypeScript
const contract = c.router(
  {
    getPost: {
      method: "GET",
      path: "/posts/:id",
      responses: {
        200: c.type<{ id: number }>(),
      },
    },
  },
  {
    metadata: SetApiRouteMetadata({ requireAuth: true }),
  }
);
```

The above solution works. But having a way to force the metadata's shape at a global level to strict what a route's metadata can be would make maintenance easier, IMO.

Gabrola:
@iceStorm that would be pretty hard to do without extensive refactoring. Module augmentation using `declare module` wouldn't work because it can only add properties to interfaces, but not change them. Can't think of a better approach than the one you already presented.

```typescript
export const SetApiRouteMetadata = <const T extends ApiRouteMetadata>(metadata: T) => metadata;
```

Doing this should be better, because the individual routes will have literal types such as `true` or `false` rather than just `boolean`.

iceStorm:
> @iceStorm that would be pretty hard to do without extensive refactoring. Module augmentation using `declare module` wouldn't work because it can only add properties to interfaces, but not change them. Can't think of a better approach than the one you already presented.
> 
> ```ts
> export const SetApiRouteMetadata = <const T extends ApiRouteMetadata>(metadata: T) => metadata;
> ```
> 
> Doing this should be better, because the individual routes will have literal types such as `true` or `false` rather than just `boolean`.

Wow, I was curious about how to make the data type from `SetApiRouteMetadata` a const. Thank you. Learned a new thing 😃

PR #586, opened 2024-04-17
docs: clarify file upload openapi

nightspite:
I have lost some time looking for a way to extend the schema to handle file upload properly - correct openapi representation was worth more for me than the typed input. 
In swagger this field is displayed as a file input

https://swagger.io/docs/specification/describing-request-body/file-upload/

CleanShot 2024-04-17 at 13 34 23@2x

### DIFF
--- a/apps/docs/docs/core/form-data.md
+++ b/apps/docs/docs/core/form-data.md
@@ -10,6 +10,9 @@ The contract implementation is the same as any other mutation, however, `content
 // contract.ts
 
 import { initContract } from '@ts-rest/core';
+import { extendZodWithOpenApi } from '@anatine/zod-openapi';
+
+extendZodWithOpenApi(z);
 
 const c = initContract();
 
@@ -18,7 +21,12 @@ export const postsContract = c.router({
     method: 'POST',
     path: '/posts/:id/thumbnail',
     contentType: 'multipart/form-data', // <- Only difference
+    // if you want to have the body typed
     body: c.type<{ thumbnail: File }>(), // <- Use File type in here
+    // if you care about openapi schema
+    body: z.object({
+      thumbnail: z.any().openapi({ type: 'string', format: 'binary'}),
+    }),
     responses: {
       200: z.object({
         uploadedFile: z.object({


Gabrola:
Unfortunately, this would break the typing for both the client and server, and also fail validation on the server. Maybe this works for you in a specific use-case, but we cannot include this in the official docs.

nightspite:
How will that fail a validation on the server? 
Current version in the docs specifies type only, no validation there.
```ts
contentType: 'multipart/form-data', // <- Only difference
body: c.type<{ thumbnail: File }>(), // <- Use File type in here
```

And as far as I remember the multipart content type is not respected in openapi schema.

nightspite:
> @nightspite, the idea is failing server validation for me when using the z.object()
> 
> Have you figured out a workaround to getting openapi to know this endpoint is a file upload?

I use something like that:
File schema
```ts
import { extendZodWithOpenApi } from "@anatine/zod-openapi";

extendZodWithOpenApi(z);

const fileUploadSchema = z.object({
  teamId: z.string(),
  file: z.any().openapi({ type: "string", format: "binary" }),
});
export type FileUpload = z.infer<typeof fileUploadSchema>;
```

Contract
```ts
create: {
  method: "POST",
  path: "/",
  description: "Upload a file. This is the only endpoint that uses multipart/form-data.",
  contentType: "multipart/form-data",
  body: fileUploadSchema,
  responses: {
    ...errorResponseObject,
    200: uploadForPublicDtoSchema,
  },
},
```

Frontend hook
```ts
export const useCreateUpload = (options?: UseMutationOptions<typeof privateContract.upload.create, any>) => {
  const queryClient = useQueryClient();

  return API_CLIENT.upload.create.useMutation({
    ...options,
    onSuccess: (data, variables, context) => {
      toast.success("File uploaded");

      queryClient.invalidateQueries([QUERY_KEYS.UPLOAD_LIST]);
      options?.onSuccess?.(data, variables, context);
    },
    onError: (error, variables, context) => {
      toastApiError("Could not create an upload.", error);

      options?.onError?.(error, variables, context);
    },
  });
};
```

Use of the hook
```ts
const formData = new FormData();

  formData.append("teamId", teamId);
  formData.append("file", file);

  await createUpload(
    {
      body: formData,
    },
    {
      onSuccess: (data) => {
        if (onUpload && data?.body) {
          onUpload(data?.body);
        }
      },
    },
  );
```

And I still get the benefit of having a correct swagger ui
CleanShot 2024-09-22 at 01 40 55@2x

PR #591, opened 2024-04-23
feat(fastify): pass appRoute inside route handler args

dmstoykov:


libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
```diff
@@ -31,6 +31,7 @@ type AppRouteImplementation<T extends AppRoute> = (
   input: ServerInferRequest<T, fastify.FastifyRequest['headers']> & {
     request: fastify.FastifyRequest;
     reply: fastify.FastifyReply;
+    schema: T;
```
Gabrola, line 34: We call this `appRoute`, so lets keep it like that for consistency's sake.

Gabrola (changes requested):

PR #595, opened 2024-04-30
Validate multipart body with @ts-rest/fastify

mbenedettini:
I was trying to get `@ts-rest/fastify` to validate a multipart body and ended up making this modification for it to happen. 

Am I in the right direction? If so, I'm happy to fix types and add tests to make this a proper contribution.

If not, can you point me how I should address validation for a scenario like this where I have a contract as follows and I'm also using `@fastify/multipart` to get streams for uploaded files? I couldn't find any other place where to hook up such validation:

```typescript
import { MultipartFile as MultipartFileFastify } from "@fastify/multipart";
const MultipartFile: z.ZodType<MultipartFileFastify> = z.any();
const Schema = z.object({
  from: z.string(),
  to: z.string(),
  text: z.string().optional(),
  subject: z.string().optional(),
  message: z.string().optional(),
  attachmentStreams: z.array(MultipartFile).optional(),
});
```

Thanks!
Mariano.

### DIFF
--- a/libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
+++ b/libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
@@ -61,6 +61,45 @@ type RegisterRouterOptions = {
       ) => void);
 };
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const getMultipartBody = (requestBody: any) => {
+  const reqBodyKeys = Object.keys(requestBody);
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const body: Record<string, any> = {};
+  for (let i = 0; i < reqBodyKeys.length; ++i) {
+    const key = reqBodyKeys[i];
+    const field = requestBody[key];
+
+    if (field.value !== undefined) {
+      body[key] = field.value;
+    } else if (field.file !== undefined) {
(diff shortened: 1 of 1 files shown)


Gabrola:
Just use `attachFieldsToBody: 'keyValues'` when registering `@fastify/multipart`

mbenedettini:
It doesn't work for the reason I submitted this PR: body values are no longer the values itself but objects from which you need to read the key `value`.

Here's a minimal proof of concept that reproduces the issue: https://github.com/mbenedettini/ts-rest-fastify-multipart-proof

Gabrola:
@mbenedettini It's not validating because you are using a pure typescript type for your schema, not a Zod schema. `attachFieldsToBody: 'keyValues'` will set the values directly in `body`, not in `.value` fields.

If you set the body to this it should work.

```typescript
z.object({
  requiredField: z.string(),
  file: z.instanceof(Buffer),
})
```

mbenedettini:
@Gabrola thanks, that works. For the record, this is how I defined the contract:

```typescript
import { MultipartFile } from "@fastify/multipart";
const Multipart: z.ZodType<MultipartFile> = z.any();
const BodySchema = z.object({
  requiredField: z.string(),
  file: Multipart,
});
```

PR #609, opened 2024-06-05
feat: adopt useSuspenseQuery

ChiefORZ:
To provide a smoother transition to the v5 of `@tanstack/query`, we should expose the new methods `useSuspenseQuery`, `useSuspenseQueries` as well as `useSuspenseInfiniteQuery`. We will fall back to the v4 default way by passing `{ suspense: true }` to the queryOptions.

This will be another push to get the stone rolling, because the new **hydration/dehydration** features are awesome.

### DIFF
--- a/libs/ts-rest/react-query/package.json
+++ b/libs/ts-rest/react-query/package.json
@@ -22,7 +22,7 @@
   },
   "peerDependencies": {
     "react": "^16.8.0 || ^17.0.0 || ^18.0.0",
-    "@tanstack/react-query": "^4.0.0",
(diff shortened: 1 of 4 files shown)


Gabrola:
@ChiefORZ Thank you for the contribution! I will close this for now, as adding react-query v5 to the `peerDependencies` causes (and has previously caused) a lot of compatibility issues and causes package managers to inadvertently install v5. As a result, we cannot merge any v5 compatible code until we actually do the transition.
Once we migrate and are ready to release ts-rest v4, I will re-open this PR. Thank you again!

PR #615, opened 2024-06-26
feat: add support for Azure Functions V4 serverless runtime

Liamandrew:
We recently did a PoC to run an app with ts-rest on Azure Functions and as part of that I did some work to map the handlers correctly based on the Azure Function V4 runtime. I figured it would be beneficial to the community to have that included.


Gabrola (approved):
This is awesome, thank you! I've pushed some fixes and will merge.

PR #620, opened 2024-06-30
Fix error Ts(4023) Update dsl.ts 

moharambe1:
fix error Exported variable 'CONTRACT' has or is using name 'tag' from external module "@ts-rest/core/src/lib/type-utils" but cannot be named.


Gabrola:
The error in your description is unrelated to what you are trying to export. Please submit an issue with a full code reproduction.

PR #621, opened 2024-06-30
Fix error Ts(2742) Update vue-query.ts

moharambe1:
The inferred type of 'client ' cannot be named without a reference to '@ts-rest/vue-query/src/lib/use-mutation'. This is likely not portable. A type annotation is necessary.


Gabrola:
You need to manually annotate your exported client

```typescript
const options = {
  baseUrl: 'http://localhost:3000',
} satisfies Parameters<typeof initQueryClient>[1];

export const client: InitClientReturn<typeof contract, typeof options> = initQueryClient(contract, options);
```

PR #624, opened 2024-07-04
feat: add options for fastify route level hooks

aabegg:
It is now possible to define fastify hooks at route level using the same principle as with ts-rest express.

```
const router = s.router(apiBlog, {
  // ...
  createPost: {
    preHandler: (request, reply, done) => {
      // some code
      console.log((request.routeConfig as { appRoute: AppRoute }).appRoute);
      done();
    },
    handler: async ({ body }) => {
      const post = mockPostFixtureFactory(body);

      return {
        status: 201,
        body: post,
      };
    },
  },
//...
});
```

All hooks that are currently possible with fastify are supported.

In addition, the contract of the respective route can be accessed in a hook with `(request.routeConfig as {appRoute: AppRoute}).appRoute`. ***This is also possible in the global hooks which are registered directly via fastify.***


Gabrola:
Thank you @aabegg, seems like this was already implemented in #265 with the additional of global hooks to, so will clean that up and merge that one instead.

PR #625, opened 2024-07-07
Add ts-rest metadata to Fastify Route Config

bestickley:
Hi, I want to be able to access route `metadata` within Fastify Route's config object so that I can do authorization in global plugin/middleware that reads permission data per route. Could you please consider this PR? The change is very simple.
cc: @oliverbutler


Gabrola (approved):
Thank you for this! I changed it to include the entire route in the config so it's more consistent with how it's done in the express library.

PR #649, opened 2024-07-29
docs: fix custom http client examples

TotallyBiased:
Minor copy-pasta error in the docs, `response` isn't defined.

It might need a cast like the `Method` above, 🤷


Gabrola:
Thank you!

Issue #636, opened 2024-07-22
@ts-rest/fastify move validation away from handler into the validation and serealization of fastify


rkreienbuehl:
**Is your feature request related to a problem? Please describe.**
Validation with `@ts-rest/fastify` is done within the handler. This makes the whole fastify hook system more or less useless. 

**Describe the solution you'd like**
It would be better and cleaner if fastify would add a type-provider for fastify and register a `validationCompiler` with `app.setValidatorCompiler(validatorCompiler)` and a `serializerCompiler` with `app.setSerializerCompiler(serializerCompiler)`.

**Describe alternatives you've considered**
Instead of writing an own type-provider, maybe [fastify-type-provider-zod](https://github.com/turkerdev/fastify-type-provider-zod) could be used.

**Additional context**
In a first test I used `fastify-type-provider-zod` to move the validation into `validate` and `serealize` into the fastify mechanism. This could be improved to fully support `ts-rest` by writing an own type-provider similar to `fastify-type-provider-zod`.

The changed parts of my code (@ts-rest/fastify `v3.46.0`) look like this:
```ts
...

export const initServer = () => (
  ...
  ) => {
    // setup type-provider
    app.setValidatorCompiler(validatorCompiler);
    app.setSerializerCompiler(serializerCompiler);

    app.withTypeProvider<ZodTypeProvider>();

    ...
  },
  plugin: <T extends AppRouter>(
    router: RecursiveRouterObj<T>
  ): FastifyPluginCallback<RegisterRouterOptions<T>> => (
    app,
    opts = {
      logInitialization: true,
      jsonQuery: false,
      responseValidation: false,
      requestValidationErrorHandler: 'combined',
    },
    done
  ) => {
    // setup type-provider
    app.setValidatorCompiler(validatorCompiler);
    app.setSerializerCompiler(serializerCompiler);

    app.withTypeProvider<ZodTypeProvider>();

    ...
  },
});

...

/**
 * @param routeImpl - User's implementation of the route
 * @param appRoute - the `ts-rest` contract for this route (e.g. with Path, params, query, body, etc.)
 * @param fastify - the fastify instance to register the route on
 * @param options - options for the routers
 */
const registerRoute = <TAppRoute extends AppRoute>(
  routeImpl: AppRouteImplementationOrOptions<AppRoute>,
  appRoute: TAppRoute,
  app: FastifyInstance,
  options: BaseRegisterRouterOptions
) => {
  
  ...

  // construct schema
[...]

PR #659, opened 2024-07-31
feat(fastify): move validation into a validatorCompiler and a serializerCompiler

rkreienbuehl:
Moving validation into a validatorCompiler and a serializerCompiler according to issue #636

### DIFF
--- a/libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
+++ b/libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
@@ -1,39 +1,29 @@
 import {
   AppRoute,
   AppRouteMutation,
-  AppRouteQuery,
   AppRouter,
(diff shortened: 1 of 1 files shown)


Gabrola:
Thank you for your PR! However, there are a couple of issues with this solution:
- Does not handle `jsonQuery`
- The new request validation error format is not backwards compatible and will break existing apps
- `passThroughExtraKeys` should not be `true` for all parts of a request.
- I don't think `serializerCompiler` is intended to be doing validation

The main issue is the problem with the request validation error, and I don't know how we can really make sure we can validate all parts of the request with the current method, so we can send all validation errors at the same time in the same format as before. I don't this is going to be possible using `setValidatorCompiler`.

I'll close this PR for now, until we can come up with a solution that can achieve this.

PR #666, opened 2024-08-09
Next getMetadata query example update

shawnmclean:
Enriching the next example with how to use generateMetadata.

The tanstack guys seems to think fetchQuery is ok: https://github.com/TanStack/query/discussions/7313

Questions:

In their example, they have the queryClient as a global, in our example we're getting it twice. Is there a way to clean this up?

### DIFF
--- a/apps/example-next/app/(react-query-streaming)/test-streaming/page.tsx
+++ b/apps/example-next/app/(react-query-streaming)/test-streaming/page.tsx
@@ -13,7 +13,7 @@ export default function Test() {
   tsrQueryClient.test.prefetchQuery({
     queryKey: ['TEST_SUSPENSE'],
     queryData: {
-      params: { id: 2 },
+      params: { id: 1 },
       query: { foo: 'test', bar: 123 },
     },
   });


apps/example-next/app/(react-query-streaming)/test-streaming/page.tsx
```diff
@@ -3,19 +3,33 @@ import { HydrationBoundary, dehydrate } from '@tanstack/react-query';
...
+import type { Metadata } from 'next';
 
 export const dynamic = 'force-dynamic';
 
+const queryData = {
+  params: { id: 2 },
+  query: { foo: 'test', bar: 123 },
+};
+
+export async function generateMetadata(): Promise<Metadata> {
```
Gabrola, line 15: It's better to have one example use-case per page. It will make it easier to find, and prevent any inadvertent confusion that the two use-cases are tied to each other, if someone is not well-versed enough in Next. In this case, someone might think that streaming and metadata are tied to each other.

We are also running the same query twice once in `generateMetadata()` and once in `Test()` but on two separate query clients, so the query will actually run twice, but the underlying `fetch` request will end up being deduplicated by Next.js pre-v15, and on v15+ it will actually run two fetches. This could be intended or it could be not, but it's better to make sure our example is unopinionated as there are different strategies of either using separate query clients, or using a cached query client per-request.

So it would be much better and avoid any potential confusion, if we have a separate example in something like `app/(react-query)/test-metadata/page.tsx` where we make sure we do not run the same query twice in the server component itself and the `generateMetadata()` function.

apps/example-next/app/api/app-router/[...ts-rest]/route.ts
```diff
@@ -1,14 +1,17 @@
...
 import { testContract } from '../../../../contracts/test-contract';
+import { randomInt } from 'crypto';
 
 const handler = createNextHandler(
   testContract,
   {
     test: async ({ params, query }) => {
+      console.log('hitting test endpoint', params, query);
+
       return {
```
Comment (Gabrola) [inline apps/example-next/app/api/app-router/[...ts-rest]/route.ts:11]:
```suggestion
      return {
```

apps/example-next/app/api/app-router/[...ts-rest]/route.ts
```diff
@@ -1,14 +1,17 @@
 import { createNextHandler } from '@ts-rest/serverless/next';
 import { testContract } from '../../../../contracts/test-contract';
+import { randomInt } from 'crypto';
```
Comment (Gabrola) [inline apps/example-next/app/api/app-router/[...ts-rest]/route.ts:3]: Node.js modules are not supported in the edge runtime

apps/example-next/app/api/app-router/[...ts-rest]/route.ts
```diff
@@ -1,14 +1,17 @@
...
   testContract,
   {
     test: async ({ params, query }) => {
+      console.log('hitting test endpoint', params, query);
+
       return {
         status: 200,
         body: {
-          ...params,
+          id: randomInt(1000),
```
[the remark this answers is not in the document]
Comment (Gabrola) [reply apps/example-next/app/api/app-router/[...ts-rest]/route.ts:14]: In server components, Next.js caches all `fetch` requests by default pre-v15. In all other cases, there is no implicit caching that will ever occur.

If you want do the random thing in your code to test this out sure, but no point in doing this in our official examples.

Gabrola (changes requested):

apps/example-next/app/(react-query-streaming)/test-metadata/page.tsx
```diff
@@ -0,0 +1,44 @@
...
+import type { Metadata } from 'next';
+
+export const dynamic = 'force-dynamic';
+
+const queryData = {
+  params: { id: 2 },
+  query: { foo: 'test', bar: 123 },
+};
+
+const tsrQueryClient = tsr.initQueryClient(getQueryClientRsc(true));
```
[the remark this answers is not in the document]
Gabrola: This query client is going to be shared for all requests. Just better move this into `generateMetadata` and don't do any fetches inside `Test()`. This example should solely cover the `generateMetadata` usage, so you can just return some static data inside the `Test` component.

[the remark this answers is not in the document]
Gabrola: See why I do not want to be opinionated on this issue in an official example https://tanstack.com/query/latest/docs/framework/react/guides/advanced-ssr#alternative-use-a-single-queryclient-for-prefetching

apps/example-next/app/(react-query-streaming)/test-streaming/page.tsx
Gabrola: https://nextjs.org/blog/next-15-rc#caching-updates

And in the official docs https://nextjs.org/docs/app/api-reference/functions/fetch#optionscache:~:text=Good%20to%20know,in%20Next.js.

There's two things at play here.

React itself is memoizing `fetch` requests in RSC across multiple `fetches` during the same request using `React.cache`. This is the behavior in React 18, and is going to be [changed in React 19](https://togithub.com/facebook/react/pull/28896) where React will not patch the `fetch` function at all. This kind of caching shouldn't cause problems anyway as it is only memoized per-request and is not persisted. You most likely really want to only call `fetch` once for any duplicated fetches on the same page. This is the problem that react-query itself has always fixed since it maintains its own cache, and does not refetch within `staleTime`. So in React 19 and/or Next.js 15, you will need to actually make sure you use a [single `QueryClient` instance](https://tanstack.com/query/latest/docs/framework/react/guides/advanced-ssr#alternative-use-a-single-queryclient-for-prefetching) across multiple server components and functions like `generateMetadata` to make sure you do not actually fetch multiple times. This is even mentioned in the RQ docs.

> Next.js already dedupes requests that utilize fetch(), but if you are using something else in your queryFn, or if you use a framework that does not dedupe these requests automatically, using a single queryClient as described above might make sense, despite the duplicated serialization.

Next.js 15 and React 19 have still not been fully released, so probably why the RQ docs have not been updated to reflect the change in this caching/memoization behavior.

The one probably causing you issues is the Next.js persistent data cache. Next.js will actually cache fetches across multiple requests and even deployments. This is explained here https://nextjs.org/docs/app/building-your-application/caching#data-cache

You can control this behavior using the [`fetchOptions.cache`](https://github.com/ts-rest/ts-rest/blob/main/apps/example-next/app/(react-query-streaming)/test-streaming/client-component-suspense.tsx#L12) option in ts-rest clients for single requests or change the default behavior for any fetches across a Next.js route segment using the [fetchCache](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#fetchcache) segment config

Gabrola:
> What I couldn't figure out, which would apply to the other pages, is how to handle a 404 not found.

@shawnmclean error handling is an opinionated issue as well. I assume you are talking about 404 errors from the API side, right? In this case, `prefetchQuery` discards errors, therefore the `queryClient` will just be empty and the client will attempt to fetch itself and handle the error itself, or in case you use `fetchQuery` in RSC such as in `generateMetadata()`, `fetchQuery` will just throw in case of any errors, and how you handle it depends on your use-case. If the data was essential for the page, then you most likely want to error out the entire page by putting the `fetchQuery` call inside a `try/catch` and then calling maybe `notFound()`

PR #667, opened 2024-08-11
fix: add OPTIONS export to nextjs app router example

tefkah:
Without `export { handler as  OPTIONS }`, `ts-rest` will not be able to handle `OPTIONS` requests, and as such will e.g. not set CORS even if enabled.


Gabrola:
@tefkah thank you!

PR #681, opened 2024-08-21
fix(serverless): update utils to be compatible with node runtime

### LINKED ISSUE #678: @ts-rest/serverless: createLambdaHandler throws "FileReader is not defined"
### Describe the bug

Attempting to use createLambdaHandler from `@ts-rest/serverless` on AWS Lamdba or in local node environment results in the following error:

> ReferenceError: FileReader is not defined

FileReader is being used in [these utils](https://github.com/ts-rest/ts-rest/blob/main/libs/ts-rest/serverless/src/lib/utils.ts) to transform blobs and arrayBuffers.
FileReader is part of the dom library and isn't available in a node environment.

Not sure if I'm missing something to get this working?

### How to reproduce

1. Create a serverless or SST framework demo.
2. Add dependencies `npm install @ts-rest/core zod @ts-rest/serverless`
3. Create a API gateway http event handler function that uses `createLambdaHandler`
4. Deploy to AWS Lambda
5. Call the route triggering the lambda
6. Observe "FileReader is not defined" in logs.

### Expected behavior

_No response_

### Code reproduction

I've been mainly testing using the serverless framework, but saw that others claimed to be using `createLambdaHandler` with SST so I created a reproduction of this error in SST to observe if that made any difference, the same error occurs there.

https://github.com/lukeggchapman/sst-ts-rest-serverless-demo

Specifically the code for the handler is here:

https://github.com/lukeggchapman/sst-ts-rest-serverless-demo/blob/main/packages/functions/src/test.ts

### ts-rest version

```
"@ts-rest/core": "^3.50.0",
"@ts-rest/serverless": "^3.50.0",
```

lukeggchapman:
Attempt to resolve #678 

Rewrite `@ts-rest/serverless/src/lib/utils.ts` to be compatible with both Node JS v18+ and WinterCG Minimum Common Web Platform API.

**FileReader** does not exist in Node and **Buffer** does not exist in Web API so the new implementation is using **Blob** and **btoa**.

**jsdom** [doesn't have a fully implemented Blob interface](https://github.com/jsdom/jsdom/issues/2555), so I've switched the jest environment to "node" for the **@ts-rest/serverless** library. This changed the way `Headers.forEach` works in that each header has it's own entry even for matching keys, so there is no longer a need to `splitCookiesString`.

Alternatively, if changing the test environment is not desirable I can switch back to **jsdom** on request and polyfill **Blob** and investigate the Headers discrepancies further.

🤔 It does seem like we should split the libraries for those that are intended for node from those for miniflare, that will require further discussion and is beyond my understanding of the use cases at this stage.

I'm not using Cloudflare workers to be able to test, however the tests in example-cloudflare-worker pass and I ran the library tests using [vitest-environment-miniflare](https://legacy.miniflare.dev/testing/vitest) locally to see if the behaviour of the Headers was consistent with node. I didn't leave this as the test environment though because there were failing tests in `ts-rest-azure-function.spec.ts` and `ts-rest-lambda.spec.ts`.

This is my first encounter with ts-rest, so hopefully I've not misunderstood intended use cases for this code.
I'll attempt to run this on AWS Lambda tomorrow and update if there are any failings.

### DIFF
--- a/.changeset/orange-frogs-unite.md
+++ b/.changeset/orange-frogs-unite.md
@@ -2,4 +2,4 @@
 '@ts-rest/serverless': patch
 ---
 
-Refactor utils to not use `FileReader` allowing `createLambdaHandler` to run in AWS Lambda node environments.
+Fix AWS Lambda handler not working in Node.js environments.
--- a/.changeset/orange-frogs-unite.md
+++ b/.changeset/orange-frogs-unite.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/serverless': patch
+---
+
+Refactor utils to not use `FileReader` allowing `createLambdaHandler` to run in AWS Lambda node environments.
--- a/libs/ts-rest/serverless/src/lib/handlers/ts-rest-lambda.spec.ts
+++ b/libs/ts-rest/serverless/src/lib/handlers/ts-rest-lambda.spec.ts
@@ -158,7 +158,10 @@ describe('tsRestLambda', () => {
           },
         };
       },
-      ping: async ({ body, params }) => {
+      ping: async ({ body, params }, { responseHeaders }) => {
+        responseHeaders.append('Cache-Control', 'no-cache');
+        responseHeaders.append('Cache-Control', 'no-store');
+
         return {
           status: 200,
           body: {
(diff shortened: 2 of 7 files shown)


lukeggchapman:
I've tested this today, and while the FileReader error is fixed, there is another problem.

Itty-router [introduced cloning](https://github.com/kwhitley/itty-router/commit/74f63abb25d12666e7c05e5e82fcf7432f8dbda4) of the response in v5.0.17 which breaks any handling of TsRestResponse [after the fetch](https://github.com/ts-rest/ts-rest/blob/main/libs/ts-rest/serverless/src/lib/handlers/ts-rest-lambda.ts#L53) as it now be the Response superclass.

Gabrola:
Thank you for this! I'll take a deep look into it as this needs to be very thorough and make sure it doesn't break any existing code.

Gabrola:
@lukeggchapman I'll fix the issue with itty-router in a separate PR. Thank you for taking this on! 🙌

PR #685, opened 2024-08-22
update react query v5 folder name from file to fix ambiguity in react native import

LeulAria:
- had issue importing react query v5 because of the confusion between the folder and the file having same name as v5 thus wans't able to import

- have made a fix that i have rename the folder lib to be v5-lib


Gabrola:
The file is not conflicting with the directory. Metro is unable to resolve because it is not reading the `exports` field in package.json
The package does not have a `v5.js` file so Metro can't resolve it; there is a `v5.cjs.js` file and a `v5.esm.mjs` file. You have multiple solutions for this problem.

```typescript
// this works because the types file is called v5.cjs.d.ts
import { initTsrReactQuery } from '@ts-rest/react-query/v5.cjs';
```

Or you can setup an alias in the metro config

```typescript
const aliases = {
  '@ts-rest/react-query/v5': '@ts-rest/react-query/v5.esm.mjs',
  // or cjs, whichiever you prefer
  '@ts-rest/react-query/v5': '@ts-rest/react-query/v5.cjs',
}
config.resolver.resolveRequest = (context, moduleName, platform) => {
  return context.resolveRequest(context, aliases[moduleName] ?? moduleName, platform);
};
```

We don't have an easy fix on our end without a huge overhaul of our tooling, so just do any of the above fixes on your end,

Frexuz:
@Gabrola didnt help to import with the .cjs extension:

```
Error: Unable to resolve module @ts-rest/core from /Users/.../node_modules/@ts-rest/react-query/v5.cjs.js: @ts-rest/core could not be found within the project or in these directories:
  node_modules
   5 | var initTsrReactQuery = require('./v5/init-tsr-react-query.cjs.js');
   6 | var typeGuards = require('./v5/type-guards.cjs.js');
>  7 | var core = require('@ts-rest/core');
     |                     ^
```

PR #688, opened 2024-08-27
Generate openapi requestBody required attribute

### LINKED ISSUE #679: Body cannot be marked as required in generated OpenAPI specification
### Describe the bug

The a non-optional `zod`-object body of a request is in the generated OpenAPI specification not marked as required.

### How to reproduce

Generate an OpenAPI specification from a contract that has route defined with a non optional `zod`-object as body and generate the specification.

### Expected behavior

The body to be marked as required.

### Code reproduction

```ts
import { generateOpenApi } from '@ts-rest/open-api';
import { z } from 'zod';

const c = initContract();

export const contract = c.router(
  {
    create: {
      path: '/collection',
      method: 'POST',
      body: z.object({field: z.string()}),
      responses: { 200: c.noBody(), },
    },
    }
);
export const openApiDocument = generateOpenApi(contract, {
  info: {
    title: 'API',
    version: '0.1.0',
  },
});
console.log(JSON.stringify(openApiDocument));
// Outputs: {"openapi":"3.0.2","paths":{"/collection":{"post":{"tags":[],"parameters":[],"requestBody":{"description":"Body","content":{"application/json":{"schema":{"type":"object","properties":{"field":{"type":"string"}},"required":["field"]}}}},"responses":{"200":{"description":"200"}}}}},"info":{"title":"API","version":"0.1.0"}}
// Should have requestBody.required set to true
```

### ts-rest version

```json
  "dependencies": {
    "@ts-rest/core": "^3.50.0",
    "@ts-rest/open-api": "^3.50.0",
    "zod": "^3.23.8"
  }
 ```

### PR DESCRIPTION
**Pull request by dv-thibeauv**
Fixes #679

### CODE REVIEW

**Comment by dv-thibeauv** [comment]
I think this is considered adding a feature? One might argue this is a bugfix, since we should"ve generated this before... If that is the case then please change the changeset 😄 .

**Comment by Gabrola** [comment]
LGTM. Thank you!

Yes, you're right. It's technically a new feature, and even a breaking change, since it will change the behavior of existing code. I can't imagine a scenario where it will actually break any existing code. I'll think about it for a bit, and if I don't see any risk in releasing as is then I'll do so. Otherwise, I'll put it behind an option flag with the current behavior being the default, and then switching the default behavior in the next major version.

**Comment by dv-thibeauv** [comment]
> Yes, you're right. It's technically a new feature, and even a breaking change, since it will change the behavior of existing code. I can't imagine a scenario where it will actually break any existing code. I'll think about it for a bit, and if I don't see any risk in releasing as is then I'll do so. Otherwise, I'll put it behind an option flag with the current behavior being the default, and then switching the default behavior in the next major version.

Actually, I was wrong... I think it's a bugfix. Consider the following minimal reproduction:
```ts
import { createExpressEndpoints, initServer } from "@ts-rest/express";
import express from "express"
import { z } from "zod";

const c = initContract();

  getPost: {
    method: 'POST',
    path: '/posts',
    body: z.object({ }),
    responses: {
      200: c.noBody() 
    },
  },
});

const app = express()

const s = initServer()
  getPost: async (b) => {
    console.log(b.body)
    return {
      status: 200,
      body: undefined
    }
  },
})

createExpressEndpoints(contract, router, app)

app.listen(6900, () => {
  console.log('Fake API is running on port 6900')
})

const resp = await fetch('http://localhost:6900/posts', {method: 'POST'})
console.log(await resp.json())
```
This will yield:
```console
[ts-rest] Initialized POST /posts
Fake API is running on port 6900
{
  issues: [
    {
      code: 'invalid_type',
      expected: 'object',
      received: 'undefined',
      path: [],
      message: 'Required'
    }
  ],
  name: 'ZodError'
}
```
So the openAPI specification does not match the runtime behavior, which seems like a bug?

dv-thibeauv:
Less changes needed now, and all tests/linting passes for me locally. I will no longer have access to this github account starting tomorrow, so any changes required will have to be made by maintainers of this project... Although given how little changes there are and that they pass the tests/linting, I think no more changes are required

Gabrola:
> So the openAPI specification does not match the runtime behavior, which seems like a bug?
@dv-thibeauv yes it's a bug, but the fix is not backwards compatible, as some might have already dealt with this spec-deviation in their own ways that I can't predict, neither can we predict in an absolute sense that this fix would not break any existing code. Rather than bumping the major version, bumping the minor version seems like a good compromise for something unlikely to break any code. SemVer is pretty flawed, so no point in sticking to it religiously and it's better to treat it as a bit of an art rather than a science.

> Less changes needed now, and all tests/linting passes for me locally. I will no longer have access to this github account starting tomorrow, so any changes required will have to be made by maintainers of this project... Although given how little changes there are, and the fact that they pass the tests/linting, I think/hope that no more changes are required

You actually made the exact changes I wanted, so will merge soon. Thank you!

QuentinRoy:
If anyone is looking for a workaround, this is what i'm currently using while waiting for this to be merged:
```ts
const openApiDocument = generateOpenApi(
  contract,
  { info: { title: 'Example', version: '3.0.0' } },
  {
    operationMapper(operation, appRoute) {
      if (
        'body' in appRoute &&
        appRoute.body instanceof ZodObject &&
        appRoute.body.isOptional() &&
        // It should always be the case at this point but TS won't be happy without this.
        operation.requestBody != null 
      ) {
        return {
          ...operation,
          requestBody: { ...operation.requestBody, required: true },
        };
      }
      return operation;
    },
  },
);
```

PR #694, opened 2024-09-10
feat(client): add response format options

aidant:
When migrating to using ts-rest as the contract definitions and api client there is a lot of added boilerplate in comparison to client libraries that return the body and throw on non success response codes.

To simplify these use cases without impeding the utility of errors as values and having strictly typed response codes, I've added support for two new client options;
- `simpleResponse` option changes the behaviour to return only the body of the request as well as reflecting this in the types
- `throwOnErrorStatus` option changes the behaviour to throw on non success responses and updates the typing to reflect this

When these options are used in combination with a contract that defines basic responses e.g. `responses: { 200: c.type<User>() }` the ts-rest client can be used like a simple RPC layer e.g. `const user = await client.getUser({ params: { id } })`

### DIFF
--- a/libs/ts-rest/core/src/lib/client-response-error.ts
+++ b/libs/ts-rest/core/src/lib/client-response-error.ts
@@ -1,0 +1,20 @@
+import { ClientArgs } from './client';
+import { AppRoute } from './dsl';
+import { SuccessfulHttpStatusCodes } from './status-codes';
+
+export class TsRestClientResponseError<
+  TAppRoute extends AppRoute = AppRoute,
+  TClientArgs extends ClientArgs = ClientArgs,
+> extends Error {
+  constructor(
+    public appRoute: TAppRoute,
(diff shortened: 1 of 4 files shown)


oliverbutler (commented):
Hi @aidant!

Thanks for the PR, I've looked over this, I like the concept!

It's nice that for trivial scenarios with the core client we no longer need to care about guarding against the != 2XX case.

My only hesitancy is that this introduces a paradigm shift for ts-rest, from errors as values to throwing errors

Personally, I like that as some of my personal views on handling JS errors as values vs throwing has shifting over the years - It's likely too late in the ts-rest lifecycle to introduce such a shift.

If we do this we've got a few other things to consier:

1. Other libs like react-query, vue-query, solid-query - should they now support the same ability
2. Dealing with 400's now requires try/catch which is not a existing ts-rest pattern

I'm going to err on the side of caution, citing our current goals of stability and LTS.

(If this gets substantial community support lets revisit this!!)

PR #702, opened 2024-09-30
feat: use fastify v5

vafanassieff:
Hey there,

All tests are passing locally.
Please let me know if you need more from me.

Best.

### DIFF
--- a/libs/ts-rest/fastify/package.json
+++ b/libs/ts-rest/fastify/package.json
@@ -8,7 +8,7 @@
   },
   "license": "MIT",
   "peerDependencies": {
-    "fastify": "^4.0.0",
+    "fastify": "^5.0.0",
     "zod": "^3.22.3"
   },
   "peerDependenciesMeta": {
--- a/libs/ts-rest/fastify/src/lib/ts-rest-fastify.spec.ts
+++ b/libs/ts-rest/fastify/src/lib/ts-rest-fastify.spec.ts
@@ -73,9 +73,9 @@ describe('ts-rest-fastify', () => {
 
   const router = s.router(contract, {
     test: async ({ request, reply }) => {
-      expect(request.routeConfig.tsRestRoute).toEqual(contract.test);
       expect(request.routeOptions.config.tsRestRoute).toEqual(contract.test);
-      expect(reply.context.config.tsRestRoute).toEqual(contract.test);
+      expect(request.routeOptions.config.tsRestRoute).toEqual(contract.test);
+      expect(reply.routeOptions.config.tsRestRoute).toEqual(contract.test);
 
       return {
         status: 200,
--- a/libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
+++ b/libs/ts-rest/fastify/src/lib/ts-rest-fastify.ts
@@ -49,10 +49,10 @@ type AppRouteImplementation<T extends AppRoute> = (
       FastifyContextConfig<T>
     >;
     reply: fastify.FastifyReply<
+      fastify.RequestGenericInterface,
       fastify.RawServerDefault,
       fastify.RawRequestDefaultExpression,
       fastify.RawReplyDefaultExpression,
-      fastify.RouteGenericInterface,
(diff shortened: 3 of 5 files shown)


this-is-tobi:
Hey, any news about this one ?

this-is-tobi:
Hey @oliverbutler, sorry to bother you and thanks for this awesome repository.

Can you give us some information about the maintenance / evolution of `ts-rest`, it seems like it's been almost 5 months since there has been a new version, are there any reasons for this ? Will it continue like this ? Should we consider making an official fork for the repository ?

In advance thanks for your answers, I wish you all the best for your projects 🪻

oliverbutler (changes requested):
Hi @vafanassieff! Thanks for the PR.

Unfortunately due to the breaking changes in fastify v5, it seems like this would be a major version change to @ts-rest.

This may be a case of needing a @ts-rest/fastify v5 much like we've done for react-query v5.

What do you think?

vafanassieff:
> Hi @vafanassieff! Thanks for the PR.
> 
> Unfortunately due to the breaking changes in fastify v5, it seems like this would be a major version change to @ts-rest.
> 
> This may be a case of needing a @ts-rest/fastify v5 much like we've done for react-query v5.
> 
> What do you think?

Hey, if we follow the semver spec it's indeed a major release, only for the `@ts-rest/fastify` package.

oliverbutler:
@vafanassieff yup, however, we follow a single version policy so we have the exact same versions cross all of our packages (due to type sharing)

We'll likely have to copy/paste the fastify lib and make a v5.

I'm unsure whether our standard will be to have the v5 impl within the fastify lib e.g. `@ts-rest/fastify/v5` or whether we just add a new `@ts-rest/fastify-v5` package.

@Gabrola went with the /v5 subpath pattern with react-query so would be good to get that one agreed before moving forwards with this pr.

abdirahmn1:
any updates on this?, i was very happy fastify was supported by ts-rest and then i found this issue :joy: :joy: :joy: :joy:  

Thanks to all the maintainers :heart:

egorFiNE:
I would love for this change to land, given it's more or less trivial.

TheAsda:
Waiting for this change

tandibar:
I don’t think this needs to be a breaking change. The library remains compatible with fastify v4, so users don’t have to upgrade to fastify v5. The only required adjustment is to relax the version constraint in the `package.json` to allow `fastify >= 4.0.0`.

Or am I overlooking something?

mabilalmirza:
I just started implementing a backend with ts-rest v3.52.1 and fastify v5.4.0. Things are working fine so far (I am just in initial stages). But following pnpm warning lead me to this feature request thread.

```
 WARN  Issues with peer dependencies found
xyz-backend
└─┬ @ts-rest/fastify 3.52.1
  └── ✕ unmet peer fastify@^4.0.0: found 5.4.0
```

I have read the comments and now I am confused whether to continue with it or not.

barthofu:
Same, i didn't have any issues so far.

Maybe just relax the version constraint?

libs/ts-rest/fastify/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "license": "MIT",
   "peerDependencies": {
-    "fastify": "^4.0.0",
+    "fastify": "^5.0.0",
```
bestickley, line 11: could you change this to `"^4.0.0 || ^5.0.0"` to prevent breaking change?

msaspence:
Any chance this can get over the line 🙏

Looks like just the version constraint needs relaxing?

kibertoad:
@oliverbutler Is there anything that can be done to help push this over the finish line? I'm happy to help create separate fastify5 package if that is how is preferable to proceed.

kibertoad:
Btw, I've created an alternative PR that avoids breaking changes and just bumps the range: https://github.com/ts-rest/ts-rest/pull/850

It should work at a runtime, but produce non-ideal types for Replies. Full solution would require merging this PR in some form.

Issue #622, opened 2024-07-02
`@ts-rest/core` client does not add optional path parameter to the request URL


omerchn:
### Describe the bug

when a path in the contract has optional parameters (eg. `/applications/:id/:version?` `version` is optional) the ts-rest client does not insert them into the path when specified.

### How to reproduce

1. create a contract like this:
```ts
export const contract = c.router({
  getApplication: {
    method: 'GET',
    path: `/applications/:id/:version?`,
    pathParams: z.object({
      id: z.string(),
      version: z.string().optional(),
    }),
    responses: {
      200: c.type<any>(),
    },
  },
})
```
2. use the client to call it like this:
```ts
    const { body, status } = await client.getApplication({
      params: {
        id: 'id1',
        version: 'version',
      },
    })
```
3. observe that the URL that the client requests is `/applications/id1/` instead of `/applications/id1/version`.

### Expected behavior

the client should match the param even if it has a `?` in the contract. 
I dug into the code a bit and this is where this behavior is missing: https://github.com/ts-rest/ts-rest/blob/main/libs/ts-rest/core/src/lib/paths.ts#L56-L68

the function `insertParamsIntoPath` does not match params that have `?`.
here is a proposed fix:
( I am not a regex expert and there could be a better fix that does not require another `replace` )

```ts
export const insertParamsIntoPath = <T extends string>({
  path,
  params,
}: {
  path: T;
  params: ParamsFromUrl<T>;
}) => {
  return path
    .replace(/:([^/?]+)/g, (_, p) => {
      return (params as any)[p] || '';
    })
    .replace(/\/\//g, '/')
    .replace('?', '');
};
```

### Code reproduction

contract
```ts
export const contract = c.router({
  getApplication: {
    method: 'GET',
    path: `/applications/:id/:version?`,
    pathParams: z.object({
      id: z.string(),
      version: z.string().optional(),
    }),
    responses: {
      200: c.type<any>(),
    },
  },
})
```
client request
```ts
const { body, status } = await client.getApplication({
  params: {
    id: 'id1',
    version: 'version',
  },
})
```

### ts-rest version

latest

mgnsk:
Also having this issue. While your proposed solution works, the optional question mark could be included in the regex itself, requiring no extra replace:

```diff
   return path
-    .replace(/:([^/]+)/g, (_, p) => {
+    .replace(/:([^/?]+)\??/g, (_, p) => {
       return (params as any)[p] || '';
     })
[...]

PR #704, opened 2024-10-05
fix(core): Replace optional path params in request URL

mgnsk:
Fixes https://github.com/ts-rest/ts-rest/issues/622.

Ran the test with `nx run --skipNxCache ts-rest-core:test`.

### DIFF
--- a/libs/ts-rest/core/src/lib/paths.spec.ts
+++ b/libs/ts-rest/core/src/lib/paths.spec.ts
@@ -92,4 +92,38 @@ describe('insertParamsIntoPath', () => {
 
     expect(result).toBe('/1');
   });
+
+  it('should insert optional params into path with many params', () => {
+    const path = '/post/:id?/comments/:commentId?/:commentId2?';
+
+    const params = {
+      commentId: '2',
+      commentId2: '3',
+      id: '1',
+    };
+
+    const result = insertParamsIntoPath({ path, params });
+
+    expect(result).toBe('/post/1/comments/2/3');
+  });
+
+  it('should insert optional params into path with no params', () => {
+    const path = '/post/:id?/comments/:commentId?/:commentId2?';
+
+    const result = insertParamsIntoPath({ path, params: {} });
+
+    expect(result).toBe('/post/comments/');
+  });
+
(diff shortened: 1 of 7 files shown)


libs/ts-rest/core/src/lib/paths.spec.ts
```diff
@@ -92,4 +92,38 @@ describe('insertParamsIntoPath', () => {
...
+
+    expect(result).toBe('/post/1/comments/2/3');
+  });
+
+  it('should insert optional params into path with no params', () => {
+    const path = '/post/:id?/comments/:commentId?/:commentId2?';
+
+    const result = insertParamsIntoPath({ path, params: {} });
+
+    expect(result).toBe('/post/comments/');
```
mgnsk, line 115: Not sure how to handle the trailing slash here, both `:commentId` and `:commentId2` are missing.

The function replaces two consecutive slashes with a single one but it doesn't handle this case where more than 1 missing optional params appear in the end of the path.

Kajatin:
Why isn't this merged?

oliverbutler:
Hi @mgnsk !

Thanks for the PR

This looks super promising, I am interested to see how frameworks handle optional params as I haven't personally utilized them (at least in Nestjs)

Before merging I want to:
1. Look into your comment about two optional params in a row, how it affects the path and adapters.
2. Test this in the e2e projects (express at least)

my gut says when it's ambiguous (two optional params one after the other) things might get tricky, especially if they have differing  types.

libs/ts-rest/core/src/lib/paths.spec.ts
oliverbutler: Fixed this such that we don't have a trailing / :+1:

oliverbutler:
@Kajatin thanks for your patience.

I had some back and forth, when I started fleshing out the tests for NestJS I realised that our existing lib implementation did not play nicely with the optional params.

I waited till I had finished https://github.com/ts-rest/ts-rest/pull/750 then finished up here.

This should now all work! :crossed_fingers:

oliverbutler (approved):
LGTM! :)

PR #713, opened 2024-11-09
fix: AppRouteDeleteNoBody  should resolve to AppRouteQueryImplementation type

### LINKED ISSUE #712: `@ts-rest/express` optional body (`AppRouteDeleteNoBody`) causes type errors in route handlers
### Describe the bug

When a contract defines a DELETE endpoint with no body (#669), the corresponding request handler results in a type of `never`.

### How to reproduce

1. Define contract with DELETE endpoint and optional body
2. Define router from `@ts-rest/express` with the given contract
3. Route handler cannot be defined due to a type error.

### Expected behavior

`AppRouteImplementation` type in `@ts-rest/express` should not resolve to `never` when contract contains `AppRouteDeleteNoBody` type.

### Code reproduction

```ts
import { initServer } from "@ts-rest/express";

const c = initContract();
const s = initServer();

  deleteReport: {
    method: "DELETE",
    summary: "",
    description: "",
    path: "/measurement-verification/reports/:reportId",
    responses: {
      204: contract.noBody()
    }
  }
});

export const routes = server.router(routeContracts, {
  deleteReport: {
    middleware: [],

    // TypeScript throws an error:
    // Type '({ req, headers, params }: { req: any; headers: any; params: any; }) => Promise<{ status: number; body: null; }>' is not assignable to type 'never'.
    async handler({ req, headers, params }) {
      return {
        status: 204,
      };
    }
});
```

### ts-rest version

@ts-rest/core: 3.51.0
@ts-rest/express: 3.51.0

AndrewJo:
Fixes #712.

This is mainly a fix for the typing system and does not break existing tests which was run with `pnpm nx affected:test`.

### DIFF
--- a/.changeset/odd-cooks-leave.md
+++ b/.changeset/odd-cooks-leave.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/express': patch
+---
+
+Fix route handler type for DELETE endpoints with no body
--- a/apps/example-express/src/authenticated-app.ts
+++ b/apps/example-express/src/authenticated-app.ts
@@ -137,6 +137,12 @@ const completedRouter = s.router(apiBlog, {
       };
     },
   },
+  deleteNoBody: async () => {
+    return {
+      status: 200,
+      body: { message: 'deleted' },
+    };
+  },
   testPathParams: async ({ params }) => {
     return {
       status: 200,
--- a/apps/example-express/src/authenticated-app.ts
+++ b/apps/example-express/src/authenticated-app.ts
@@ -137,12 +137,6 @@ const completedRouter = s.router(apiBlog, {
       };
     },
   },
-  deleteNoBody: async () => {
-    return {
-      status: 200,
-      body: { message: 'deleted' },
-    };
-  },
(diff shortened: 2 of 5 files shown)


strblr:
What's the status on this? I'm encountering the same TS error when defining a DELETE without body.

AndrewJo:
@strblr The tests pass locally on my dev machine but it's up to the maintainers to approve and merge it. 🤷🏻‍♂️ 

There has been no further activities from the maintainers (both here or on the Discord server) since the last release in August which is already 5 months ago so I'm beginning to wonder if this project has been abandoned.

oliverbutler (approved):
Looks great! Good catch and nice fix.

I tested locally and added a regression test in the express example app for delete without body (confirmed it type erorrs _without_ your change)

oliverbutler:
@AndrewJo when removed `body` from the deletePost in `apiBlog` i noticed it unfortunately also broke the next example for react query v5.

I've done the same fix there.

I wasn't part of the implementation of optional body on a delete - so shall see :crossed_fingers:

PR #727, opened 2024-12-12
Grouping schemas by title and referencing them on `@ts-rest/openapi`

### LINKED ISSUE #366: feature: deduplicate schemas in generated OpenAPI
Let's say you have a Post entity that's returned from a bunch of endpoints. Repeating that entity in the OpenAPI schema at every endpoint will create a large .JSON file. It probably also makes it harder for downstream tooling to organize types (would end up with Post1 Post2 etc) - this is one advantage of schema-first codegen driven development right now, there is clean code available when you e.g. jump to definition which takes you to a nicely-named interface

Check out the OpenAPI specification's components: https://swagger.io/specification/#:~:text=Specification%20Extensions.-,Components%20Object,-Holds%20a%20set

Perhaps referential equality could be used to output refs. OR, change the library's API to allow specifying shared schemas.

EDIT: Referential equality probably wouldn't work; even if the Zod schema is referentially equal (which it may not be if the user defines it twice for some reason), `@anatine/zod-openapi` probably doesn't create referentially equal objects. The objects would have to be deduped by content (perhaps JSON.stringify if it can be trusted that the objects are stringifiable and deterministically ordered). This kind of magic seems like it'd be brittle.

nokutu:
Fixes #366

By using `@anatine/zod-openapi` and setting the title property of a zod object, we can extract the schemas into `components` and reuse them around the code.

This will greatly reduce the output size of big APIs as well as generate much more compact SDKs automatically.

It assumes titles will be unique, and in case they are not, an exception will be thrown. We could use any other variable if you think it fits better.

### DIFF
--- a/.changeset/twenty-fans-reflect.md
+++ b/.changeset/twenty-fans-reflect.md
@@ -1,0 +1,5 @@
+---
+'@ts-rest/open-api': minor
+---
+
+Support automatic reusable "Components" in OpenAPI schemas, reduces final size of schema, uses the "Title" property from @anatine/zod-openapi
--- a/libs/ts-rest/open-api/src/lib/ts-rest-open-api.spec.ts
+++ b/libs/ts-rest/open-api/src/lib/ts-rest-open-api.spec.ts
@@ -1123,7 +1123,7 @@ describe('ts-rest-open-api', () => {
 
       expect(schema).toEqual({
         components: {
-          schema: {
+          schemas: {
             Body: {
               properties: {
                 file: {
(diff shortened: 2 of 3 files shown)


strblr:
This would be really nice! What's the status on it?

nokutu:
No idea, I guess somebody from the team need to authorize Vercel deployment

oliverbutler (approved):
Great PR!

Thanks for this, this is a huge improvement to the openapi implementation.

I've fixed and import and added a minor changeset.

Great work :)

PR #750, opened 2025-02-10
feat: migrate away from `@All` for multi-handler nest routes 

oliverbutler:
Try it -> 3.52.0-alpha.0

Historically we've been stuck using "@All" for multi handler in nest.

Nest doesn't allow you to use multiple verb decorators on a single method.

This has historically come with some major problems:

### Controller collisions

If two multi handler controllers have the same path e.g. `GET /posts` in file 1, `POST /posts` in file 2 these will conflict given that they use @All under the hood, even though they should _not_ collide

We've added a warning in the docs for this

### Routing individual requests

As we decorated a single method e.g. `handle` with `@All(['/posts', '/posts/:id'])` we had to write our own small routing logic which decided which method in the multi handler to invoke

This has not (to our knowledge) caused any problems yet (taps on :wood:) but it VERY easily could, it makes sense that nest is the one in control of the routing logic to decide which method to invoke.

## Proposed solution

Virtual methods :sparkle:

> A totally official name I've pulled out of my... brain.

 Imagine we've got this pseudo controller.

```ts
// Original class
class PostController {
  async handler(contract) {
    return tsRestHandler(contract, {
      getPosts: {...},
      getPost: {...}
    });
  }
}

// At runtime it becomes:
class PostController {
  @Get("/posts")
  async handler_getPosts() {
    return this.handler(...);
  }

  @Get("/posts/:id") 
  async handler_getPost() {
    return this.handler(...);
  }

  async handler(contract) {
    return tsRestHandler(contract, {
      getPosts: {...},
      getPost: {...}
    });
  }
}
```

This means that multi handler becomes syntactic sugar for the single approach, removing the two problems we originally faced.

### DIFF
--- a/libs/ts-rest/nest/src/lib/ts-rest-nest-handler.ts
+++ b/libs/ts-rest/nest/src/lib/ts-rest-nest-handler.ts
@@ -151,7 +151,6 @@ export const TsRestHandler = (
             return originalMethod.apply(this, args);
           };
 
-          // Copy the route arguments metadata to the new method
           if (originalParamMetadata) {
             Reflect.defineMetadata(
               ROUTE_ARGS_METADATA,
@@ -161,7 +160,6 @@ export const TsRestHandler = (
             );
           }
 
-          // Get parameter types
           const paramTypes = Reflect.getMetadata(
             'design:paramtypes',
             target,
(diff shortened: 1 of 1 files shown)


oliverbutler:
Testing:

I've deployed as an alpha package, I've imported into my companies stack (3.52.0-alpha.1) and ran the full battery of builds, tests, e2e etc cross stack.

It uses @ts-rest/nest every-which possible way so I have huge faith that the solution is solid.

It did identify some issues with parameter decorators, which i've now added regression tests in @ts-rest/nest for.

Issue #743, opened 2025-02-09
Accept Standard Schemas in ts-rest


iGoodie:
**Is your feature request related to a problem? Please describe.**
No it is not a problem, rather a maintenance concern.

**Describe the solution you'd like**
Authors of most popular validation libraries like Zod, ArkType and Volibot came together to form a standard.
It would be amazing, if ts-rest could accept any Validation library that implements Standard Schema.
They briefly explain how to accept under: https://github.com/standard-schema/standard-schema#how-do-i-accept-standard-schemas-in-my-library

**Describe alternatives you've considered**
-

**Additional context**
-

aidant:
@oliverbutler I've been thinking similarly and would be willing to look at implementing this as a feature given the opportunity

oliverbutler:
Hi @aidant

I was working may way though my backlog of issues and hadn't stumbled upon this yes.

Absolutely, 100%.

Dependency on Zod is less necessary with this promising standard.

Feel free to have a stab at this, I'm more than willing to help as much as I can. 

This is likely a large change!

kevinmarrec:
@oliverbutler This feature would be amazing, this is probably the only feature I'm lacking to use `ts-rest` in production.

Here are some details :

- `ts-rest` client requires the contract to be passed as a value and not only as a type
- `zod` is awesome but bundle-size heavy for frontends because lack tree-shaking vs other solutions (e.g. `valibot`) 
- Therefore, as your contract will be shipped in the frontend, you'll have full `zod` included in the final bundle

> Nice links :
> https://bundlephobia.com/package/zod@3.24.2
> https://www.builder.io/blog/valibot-bundle-size

Proposals :
- **feat**: Supports Standard Schemas (this issue)
- **perf**: Be able to use `initClient` by only giving the type of the contract (like `trpc`, but still benefit of shared contract and not importing types from server)

LeulAria:
will their be any update on this ?

antonio-ivanovski:
Would be nice to see! Would the TS developer experience be improved performance wise if the schemas got replaced from Zod into ArkType?

aidant:
I've got a [pull request](https://github.com/ts-rest/ts-rest/pull/751) with a working implementation. You are welcome to [try it out using gitpkg](https://github.com/ts-rest/ts-rest/pull/751#issuecomment-2715765339).

oliverbutler:
Marking this as closed, done! It's going to come out of RC to stable super soon :)

PR #751, opened 2025-02-11
feat(standard-schema): support standard schema

aidant:
[Standard Schema](https://standardschema.dev/) is a new specification that is implemented by Zod and other validation libraries. I have added support for this specification while reducing the amount of breaking changes to the external facing ts-rest API. There are however breaking changes.

I've been able to fully replace Zod with Standard Schema in all but two places - `@ts-rest/core` and `@ts-rest/open-api`. The only non deprecated usage of Zod in `@ts-rest/core` is for `ValidationErrorSchema`, `RequestValidationErrorSchema`, and `RequestValidationErrorSchemaForNest`. I have however ensured that the public api for these is typed as a Standard Schema, this would allow us to replace the implementation of these schemas without a breaking change. Finally the `@ts-rest/open-api` package is fairly tightly coupled to Zod. It would be possible to support other validation libraries by following a similar pattern as [`hono-openapi`](https://github.com/rhinobase/hono-openapi). This however would be quite a substantial change and would require updating `generateOpenApi` to be asynchronous. This feels out of scope for this already large PR.

## Releasing into 3.53.0

Hey folks, we're super close on this one - we've bolstered our test suites to ensure this is truly a non breaking change and published an alpha. 

We've also found a solution for open-api such that anyone could use any validator lib they desire

Merge Checklist (@oliverbutler )
- [x] Ensure codebase works with Zod < 3.24.0 (before standard schema added)
- [x] Release preview 
- [x] Unify typing of Legacy Zod vs Standard Schema
- [x] Improve typing of the new open-api schema transformer function, get a few examples ready for the docs
- [x] Remove breaking change for response schemas
- [x] Ensure zod is import Type everywhere
- [x] Docs - follow up PR @oliverbutler


WonderPanda:
Amazing news! This will be huge for ts-rest :heart:

iGoodie:
Absolute legend! And only in 3 days after #743 ! 🎉

apps/example-azure-function/src/functions/index.spec.ts
```diff
@@ -73,6 +73,7 @@ describe('example-azure-function', () => {
       message: 'Request validation failed',
       bodyErrors: null,
       headerErrors: {
+        name: 'ValidationError',
```
Gabrola, line 76: This would be a breaking change. Would be better if we return `ZodError` when the vendor is `zod` and then `ValidationError` for the rest.

jkazimierczak: Not PR author nor maintainer, but wouldn't it increase the bundle size just for including ZorError? The promise of Standard Schema is to decouple validator-dependent libraries from validator-specific logic.

I'm not sure if this is possible, but maybe this can be worked around by dynamically importing ZorError?

jkazimierczak: Pardon, didn't notice you were referring to string, not an actual error. In that case `vendor` is a property provided by Standard Schema itself, my bad.

aidant: @Gabrola I've updated the implementation to default to the existing behaviour if the vendor is zod resulting in `ValidationError` now being a union of `ZodError` or `StandardSchemaError` (renamed from `ValidationError`).

I'm wondering if its worth introducing a new option to disable the `ZodError` and instead just use `StandardSchemaError` with the intent of defaulting the flag in a future release and removing it eventually?

aidant:
@Gabrola thanks, its been an interesting challenge. Currently I've not updated any docs - if that is not a blocker for getting a dev build out than I would be interested in getting feedback from the community.

elee1766:
encouraging people to try this out with something like 
```
    "@ts-rest/core": "https://gitpkg.vercel.app/aidant/ts-rest/libs/ts-rest/core?feat-standard-schema",
    "@ts-rest/fastify": "https://gitpkg.vercel.app/aidant/ts-rest/libs/ts-rest/fastify?feat-standard-schema",
```
this works in yarn4 at least!

i have been playing around and its been cool and working so far

imprakharshukla:
Hey great work guys!

Just wanted to ask, how close are we to merging this? Seeing lot of recent commits by @oliverbutler, makes me think we are close (at least I hope we are).

I cannot test it with gitpkg using pnpm (it does not infer the types for some reason).

Is there a way we can get actual built pre-release packages for this?

oliverbutler:
Hey all, @imprakharshukla I feel like we're quite close.

I've pushed up some more regression testing to check how non breaking we are for zod3, and check our support for other standard schema validators.

I think you're right that the remaining issue is the inferred types error from one of the libs.

Once the CI is passing I'll deploy a preview version we can get feedback on.

oliverbutler:
Just pushed up [`2ea08da` (#751)](https://github.com/ts-rest/ts-rest/pull/751/commits/2ea08dac03ecf5ea0b24c5c5a31e6830c6af2aa1), it's a large commit, admittedly. 

This brings support for open-api in a post-standard-schema world.

It follows a BYOT (bring your own transformer) mode, theres a bunch of competing libs that offer `Schema Lib -> OpenAPI` some requiring intermediaries, for instance:

Zod 3 - `@anatine/zod-openapi`
Zod 4 - `Builtin z.toJSONSchema` ➡️ `@openapi-contrib/json-schema-to-openapi-schema` 
Valibot -> `@valibot/to-json-schema` ➡️  `@openapi-contrib/json-schema-to-openapi-schema` 

This seems like a bad idea to try keep up with, we'll have a huge flow of PRs suggesting new libs we're missing, dealing with the downstreams having breaking changes - I don't want that!

So instead you can do this:

```ts
const SCHEMA_TRANSFORMER: SchemaTransformerAsync = async (schema) => {
  if (isZodObject(schema)) {
    const jsonSchema = z.toJSONSchema(schema);

    return await convert(jsonSchema);
  }

  return null;
};

 const openApiSchema = await generateOpenApiAsync(
        contract,
        {
          info: {
            title: 'Pokemon API',
            version: '1.0.0',
          },
        },
        {
          schemaTransformer: SCHEMA_TRANSFORMER,
        },
      );
```

You can bring your own schema (not final) letting you do whatever the heck you want, so long as we get a nice "SchemaObject" as the output - write your own, use a lib, whatever.

We also default to a builtin one for zod/3 so it's non breaking.

oliverbutler:
Final changes (🤞 ) pushed up - the base repo zod version is back to what it was on main (i.e. a version _before_ Zod supported standard schema)

This is to ensure we fully support older versions of Zod so this remains a non breaking change.

oliverbutler:
In hindsight this is how we should've typed path params too, most likely, allowing for us to provide type hints that your pathParams schema must be 1-1 with the variables in the path.

oliverbutler:
Latest alpha 3 seems to be in a great place now. 

I'm going to go for the merge, continue some more testing and publish a 3.53.0 RC

PR #754, opened 2025-02-12
fix(open-api): extract nested req body refs by title

lukashroch:
Small follow-up fix on https://github.com/ts-rest/ts-rest/pull/727

Current implementation doesn't extract nested refs on body as it only checks if body as title. This allows to extract and reference nested refs without have title on top-level body object.


lukashroch:
Hi @oliverbutler, any chance to have a look at this one, soon? Schema referencing is great feature but this small bit prevents to re-use them in nested request bodies.

PR #778, opened 2025-03-25
fix: url params interpolation with strict check

### LINKED ISSUE #777: Invalid path params interpolating with zero
### Describe the bug

Invalid handling path params. If you pass zero to params it is ignored and an invalid path is used.

### How to reproduce

Pass zero as path param to a query. For example, if you pass zero for the path "/entity/:id" the resulting path is going to be "/entity/".

### Expected behavior

Path should be "/entity/0"

### Code reproduction

```ts
c.router({
  entityById: {
    method: 'GET',
    path: '/entity/:id',
    pathParams: z.object({
      id: z.coerce.number(),
    }),
    responses: {
      200: c.type<{}>(),
    },
  },
});

// ...

client.entityById.useQuery({
  queryKey: ['entity', 0],
  queryData: {
    params: {
      id: 0,
    },
  },
});
```

### ts-rest version

3.52.1

TheAsda:
This fixes #777

### DIFF
--- a/libs/ts-rest/core/src/lib/paths.ts
+++ b/libs/ts-rest/core/src/lib/paths.ts
@@ -51,10 +51,10 @@ export const insertParamsIntoPath = <T extends string>({
   path: T;
   params: ParamsFromUrl<T>;
 }) => {
-  const pathParams = params as Record<string, string>;
+  const pathParams: Record<string, string> = params;
 
   return path.replace(/\/?:([^/?]+)\??/g, (matched, p) =>
-    pathParams[p] !== undefined && pathParams[p] !== null
+    pathParams[p]
       ? `${matched.startsWith('/') ? '/' : ''}${pathParams[p]}`
       : '',
   );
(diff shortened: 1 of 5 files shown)


libs/ts-rest/core/src/lib/paths.ts
```diff
@@ -54,7 +54,7 @@ export const insertParamsIntoPath = <T extends string>({
   const pathParams = params as Record<string, string>;
 
   return path.replace(/\/?:([^/?]+)\??/g, (matched, p) =>
-    pathParams[p]
+    pathParams[p] !== undefined && pathParams[p] !== null
```
vladjerca, line 57: This can also be `pathParams[p] != null`, nilCheck basically.

vladjerca (approved):

oliverbutler (approved):
Hey! Thanks for the PR - I wondered why I never ran into this bug in our codebase, it's most likely that we tend to not use IDs where 0 is possible, but also because we never used z.coerce.number() as it didn't used to exist, we used z.string().transform(Number)

I've made some changes to make this do the change at the level before this function, now theres no type errors throughout, we allow numbers/etc as first class citizens.

Issue #775, opened 2025-03-17
Support for EventSource


damieng57:
I attempted to create an event source endpoint using Express, but it appears that this functionality is not supported out-of-the-box. I encountered an infinite loop, and the message was not sent.

To address this, I added the content-type: text/event-stream header in my contract and modified the mainReqHandler in the ts-rest/express file as follows:

```
if (validationResults.headersResult.success && validationResults.headersResult.data.contentType === 'text/event-stream') {
   return;
 }
```
Before tried to do any changes, I have a couple of questions:

Is there a way to use EventSource without any modifications?
What is the recommended approach to create the EventSource connection on the frontend? Currently, I am making a request using event-source-polyfill without utilizing react-query. Do you think it's possible to do that with react/query?

Any guidance or suggestions would be greatly appreciated.

damieng57:
This fix https://github.com/ts-rest/ts-rest/pull/779

PR #779, opened 2025-03-26
Support for EventSource #775

damieng57:


damieng57:
Related to this issue: Support for EventSource #775

PR #728, opened 2024-12-13
chore: support express v5

ChristopheBougere:
Fixes #715 

Cf https://expressjs.com/en/guide/migrating-5.html

In the `@ts-rest/express` package, I allowed a second major version as peer dependency

I also upgraded the repo express version to ensure tests are still running on the new version. Not sure if I was supposed to do it.

PR #784, opened 2025-04-02
feat: @ts-rest/express-v5

nickdirienzo:
This PR builds on the work in #728 and #749 to add support for Express v5 in ts-rest.

Express v5 introduces several significant changes compared to v4. For a complete list of potential breaking changes, refer to the official [Express v5 migration guide](https://expressjs.com/en/guide/migrating-5.html).

There are two major differences ts-rest needs to handle:

1. Path route matching syntax: routes must now conform to the new v5 format (see "Path route matching syntax" in the guide above).
2. Promise handling: Express v5 automatically handles rejected promises, so middleware and handlers should return void instead of chaining responses (see "Rejected promises handled from middleware and handlers").

To make both v4 and v5 play nicely in the monorepo, I dropped express dependencies from the root package.json and instead moved them as `devDependencies` into their respective workspaces. This seems to be the pattern used by nest, so I went with it 🤷 

Similar to upgrading an app from Express v4 to v5, adopting ts-rest with Express v5 will require using the updated route syntax in the contract definition (see test examples for details). These are passed directly to the handler at runtime. From what I can tell there isn't a specific ts-rest path syntax that this needs to be support -- please let me know if I missed something.

Given the scope of changes, this PR introduces a new package: @ts-rest/express-v5. This allows consumers to explicitly opt into Express v5 support while enabling the ts-rest team to gather usage analytics across versions. It also provides a smoother transition path: Express v4 is still in maintenance mode and will receive security patches and high-priority fixes through late 2026, so we shouldn’t force users to upgrade before they’re ready (IMHO).


nickdirienzo:
I'll need to update the changeset but I wanted to wait on that until we felt really close to the finish line.

nickdirienzo:
If this plan of `@ts-rest/express-v5` sounds good to y'all @oliverbutler @Gabrola, I'll need your help in getting that over the finish line.

nickdirienzo:
Tests seem to be passing locally for both Express variants:

* v4: https://cloud.nx.app/runs/1SIBqrbzeE
* v5: https://cloud.nx.app/runs/O6XVbSQhNl

Have there been any discussions around having CI run on PRs so contributors can know build failures ahead of merging?

nickdirienzo:
Closing in favor of #786

PR #786, opened 2025-04-04
feat: add express v5 support to @ts-rest/express

Shakeskeyboarde:
This PR builds on the work in https://github.com/ts-rest/ts-rest/pull/728, https://github.com/ts-rest/ts-rest/pull/749, and https://github.com/ts-rest/ts-rest/pull/784 to add support for Express v5 in ts-rest.

Express v5 introduces several significant changes compared to v4. For a complete list of potential breaking changes, refer to the official [Express v5 migration guide](https://expressjs.com/en/guide/migrating-5.html).

There are two major differences ts-rest needs to handle:

1. Path route matching syntax: routes must now conform to the new v5 format (see "Path route matching syntax" in the guide above).
2. Promise handling: Express v5 automatically handles rejected promises, so middleware and handlers should return void instead of chaining responses (see "Rejected promises handled from middleware and handlers").

(Thanks @nickdirienzo for the above description text)

These changes pass both the express [4](https://github.com/ts-rest/ts-rest/tree/main/test-projects/express-4) and [5](https://github.com/ts-rest/ts-rest/tree/main/test-projects/express-5) e2e tests.

If you run into problems testing this in an actual express project, make sure your dependency graph only includes express 5. If both express 4 and 5 are present, then it's possible that there may be type mismatches between @ts-rest/express, 3rd party middleware, and your own handlers.


libs/ts-rest/express/src/lib/ts-rest-express.ts
```diff
@@ -130,6 +131,14 @@ const validateRequest = (
   };
 };
 
+const getExpressPath: (path: string) => string =
+  // The app.del() method exists in Express v4 but was removed in v5
+  'del' in application
+    ? // Express v4 accepts ts-rest paths without modification
+      (path) => path
+    : // Express v5 requires {/:param} instead of /:param? for optional params
```
nickdirienzo, line 139: When I first looked at the code I was confused because I thought ts-rest was sending the path directly to the integration, but in hindsight, there needs to be some ts-rest path convention that works for both clients + servers. I now understand why this switch is here. 

Separately, it would probably be good for the docs to be updated to explain a bit about how paths are supported and translated in contracts. Maybe it's there already and I completely missed it. 

Awesome that you got this to support both versions 🦸

nickdirienzo: If anyone else was confused like I was, check out https://github.com/ts-rest/ts-rest/blob/main/libs/ts-rest/core/src/lib/paths.ts 😄

libs/ts-rest/express/src/lib/ts-rest-express.ts
```diff
@@ -264,26 +279,32 @@ const requestValidationErrorHandler = (
       // old-style error handling, kept for backwards compatibility
       if (handler === 'default') {
         if (err.pathParams) {
-          return res.status(400).json(err.pathParams);
+          res.status(400).json(err.pathParams);
+          return;
```
nickdirienzo, line 283: I was nervous about removing these returns in #784 but looking at this now, the way ts-rest integrates into express, you can't even chain anything -- even if it is purely an express convention. Thanks again for explaining that in Discord 🙏

libs/ts-rest/express/src/lib/ts-rest-express.ts
oliverbutler: This is great! I'll definitely play with this locally, possibly find a way to test this - anytime I see a mysterious regex I get mildly concerned haha

libs/ts-rest/next/src/lib/ts-rest-next.ts
```diff
@@ -324,16 +324,20 @@ const handlerFactory = (
         }
 
         if (!pathParamsResult.success) {
-          return res.status(400).json(pathParamsResult.error);
+          res.status(400).json(pathParamsResult.error);
+          return;
```
oliverbutler, line 328: What does removing the returns from the next lib accomplish, sorry? I understand the rationale in the express project but less so here, consistency?

oliverbutler (approved):
This looks great! V5 compatibility without a /v5 is great news IMO.

I want to write a little more tests around this but this is very neat.

oliverbutler:
Hmm.. I've been trying to run the CI locally as I saw Vercel failed to build with this same error.

Screenshot 2025-05-22 at 21 36 18

It _seems_ changing these imports as follows fixes the build

Screenshot 2025-05-22 at 21 38 17

It also seems that the express-4 e2e project, which unlike NX's single dependency pattern _can_ try this with express 4 installed, in that project it seems fine with my change.

I'll push this up to the PR and get it passing but i'm unsure if this will cause issues, any thoughts @Shakeskeyboarde @nickdirienzo ?

Shakeskeyboarde:
@oliverbutler regarding the express-serve-static-core: I switched it to express for a couple of reasons.

First, the `express-serve-static-core` package isn't a dependency. So, it shouldn't be imported. It's type only, which means that when it's missing in a downstream package (ts-rest/express consumer), as long as you're not type checking node_modules, any type referencing it will essentially be `any` and will not cause an error, except to possibly weaken type checking in that downstream project overall.

I thought of making it another peer. But, it's not necessary (express v4 and v5 seem to have perfectly adequate typing), and it's old and unofficial.

So, my thinking is that it shouldn't be used at all, anywhere. It's outdated tech debt. If a project is using it for some reason, it will need to provide the dependency itself. But, because it's old, there's a reasonable chance at some point the express types will cease to align. But, that's not something that should be "fixed" in ts-rest or its supporting packages, because it runs the risk of making the problem worse.

oliverbutler:
Ok, thanks for clarifying - I agree wholeheartedly with the removal. 

This therefore means my latest commit to remove the final references in the codebase to be adequate as far as I am aware.

Are we happy to go for the merge? @Shakeskeyboarde

Shakeskeyboarde:
Yup, ready.

PR #795, opened 2025-04-30
fix(serverless): inferred types not portable

aidant:
When building with `verbatimModuleSyntax` (which enables `isolatedModules`) I get the following error:

```
error TS2742: The inferred type of 'Router' cannot be named without a reference to 'packages/ts-rest/node_modules/@ts-rest/serverless/src/lib/types.js'. This is likely not portable. A type annotation is necessary.
```

To solve this I have previously been deep importing types from `@ts-rest/serverless/src/lib/types` to manually annotate with the correct types. But this does not work when `moduleResolution` is set to more modern values such as `nodenext`. So to avoid this error entirely we can export the required types.

### DIFF
--- a/libs/ts-rest/serverless/src/index.ts
+++ b/libs/ts-rest/serverless/src/index.ts
@@ -1,5 +1,14 @@
 export * from './lib/http-error';
 export * from './lib/request';
 export * from './lib/response';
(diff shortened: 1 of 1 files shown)


oliverbutler (approved):
Thanks for the PR, added a changeset :)

